PR #6, opened 2022-06-10
4 initialize sarcophagus

knoll3:
## Initialize Sarcophagus
Implements `initializeSarcophagus()`.

### Notes
The main changes between `createSarcophagus` of V1 and `initializeSarcophagus` of V2 is that there are multiple archaeologists being added. The complication is that we need to store bounties, digging fees, and storage fees per archaeologist per sarcophagus. To do this, the archaeologist addresses are stored on the sarcophagus. AppStorage has a mapping called `sarcophagusArchaeologists` which is a mapping of bytes32 (sarc identifier) to a mapping of address to an `Archaeologist` struct. This struct contains the fees that need to be stored on the contract.

Example usage: 
`uint256 bounty = sarcophagusArchaeologists[identifier][archAddress];`

### Facet layout
This PR also sets up the structure of the facets for V2. The facets are split up by role. There is a contract for everything the embalmer should do, a contract for everything the archaeologist should do, and a contract for everything that anyone can do. Like this: 
```
contracts
  EmbalmerFacet.sol
    functions: initialize, finalize, rewrap, cancel
  ArchaeologistFacet.sol
    functions: addFreeBond, withdrawFreeBond, unwrap, transfer (and any functions associated with transfer)
  AnonymousFacet.sol (or something like that)
    functions: accuse, bury, cleanup
```


knoll3:
I just noticed a case where the embalmer could initialize a sarcophagus with an empty array of archaeologists. I added a fix for this in my current PR.

contracts/facets/EmbalmerFacet.sol
```diff
@@ -0,0 +1,130 @@
...
+            embalmer: msg.sender,
+            recipientAddress: recipient,
+            arweaveArchaeologist: arweaveArchaeologist,
+            archaeologists: archaeologistAddresses
+        });
+
+        // Add the identifier to necessary data structures
+        s.sarcophagusIdentifiers.push(identifier);
+        s.embalmerSarcophaguses[msg.sender].push(identifier);
+        s.recipientSarcophaguses[recipient].push(identifier);
```
sethhrbek, line 101: Did we decide to leave these storage mappings off the contract (`archaeologist/embalmer/recipient sarcphaguses`), and instead use events filtering for the front-end to display these?

contracts/facets/EmbalmerFacet.sol
```diff
@@ -0,0 +1,130 @@
...
+
+        // Calculate the total fees in sarco tokens that the contract will
+        // receive from the embalmer
+        uint256 totalFees = LibBonds.calculateTotalFees(
+            archaeologists,
+            s
+            .sarcophagusArchaeologists[identifier][arweaveArchaeologist]
+                .storageFee
```
sethhrbek, line 109: Does the contract need to do any validation around this `storageFee` amount? Maybe it doesn't, just wondering if there would be issues if it was undefined.

contracts/facets/EmbalmerFacet.sol
```diff
@@ -0,0 +1,130 @@
...
+        address arweaveArchaeologist,
+        address recipient,
+        uint256 resurrectionTime,
+        bytes32 identifier,
+        IERC20 sarcoToken,
+        bool canBeTransfered
+    ) external returns (uint256) {
+        // Confirm that this exact sarcophagus does not already exist
+        require(
+            !s.sarcophaguses[identifier].exists,
```
sethhrbek, line 37: Wondering if this should use `state` like in v1, b/c after a sarcophagus is cancelled / buried (and potentially cleaned), we would want to represent that state (so we can revert if the sarco is "done")

contracts/facets/EmbalmerFacet.sol
```diff
@@ -0,0 +1,130 @@
...
+        s.sarcophaguses[identifier] = LibTypes.Sarcophagus({
+            name: name,
+            exists: true,
+            canBeTransferred: canBeTransfered,
+            resurrectionTime: resurrectionTime,
+            arweaveTxId: "",
+            embalmer: msg.sender,
+            recipientAddress: recipient,
+            arweaveArchaeologist: arweaveArchaeologist,
+            archaeologists: archaeologistAddresses
```
sethhrbek, line 95: Do we need to store these addresses here, since they are stored in the `sarcophagusArchaeologists` mapping? Or could we store the archaeologist type array here instead of the addresses (since the type includes the address) and remove the mapping?

contracts/facets/ArchaeologistFacet.sol
```diff
@@ -0,0 +1,88 @@
...
+    function depositFreeBond(
+        address archaegologistAddress,
+        uint256 amount,
+        IERC20 sarcoToken
+    ) external {
+        // Confirm that sender is the archaeologist
+        require(
+            msg.sender == archaegologistAddress,
+            "sender must be the archaeologist"
+        );
```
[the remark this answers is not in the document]
knoll3: I'll definitely give this a try.

contracts/facets/EmbalmerFacet.sol
knoll3: It didn't seem like we came to a definite conclusion on that discussion so I left it in here. I agree though that we should try to get this information from the events.

contracts/facets/ArchaeologistFacet.sol
```diff
@@ -0,0 +1,88 @@
...
+contract ArchaeologistFacet {
+    AppStorage internal s;
+
+    /// @notice Deposits an archaeologist's free bond to the contract.
+    /// @param archaegologistAddress The address of the archaeologist whose
+    /// free bond is being deposited
+    /// @param amount The amount to deposit
+    /// @param sarcoToken the sarcophagus ERC20 token SARC
+    function depositFreeBond(
+        address archaegologistAddress,
```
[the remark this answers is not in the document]
knoll3: Fixed

contracts/facets/EmbalmerFacet.sol
```diff
@@ -0,0 +1,130 @@
...
+    /// @return The index of the new sarcophagus
+    function initializeSarcophagus(
+        string memory name,
+        LibTypes.Archaeologist[] memory archaeologists,
+        address arweaveArchaeologist,
+        address recipient,
+        uint256 resurrectionTime,
+        bytes32 identifier,
+        IERC20 sarcoToken,
+        bool canBeTransfered
```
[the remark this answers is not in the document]
knoll3: Fixed.

contracts/facets/ArchaeologistFacet.sol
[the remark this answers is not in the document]
knoll3: The answer is the output looks something like this 
```
'SenderNotArch("0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", "0x0000000000000000000000000000000000000000")'
```

contracts/facets/EmbalmerFacet.sol
knoll3: Trying to remember... Did we talk about being able to derive that state somehow? 

If not, then we should make this an enum instead of a boolean.

contracts/facets/EmbalmerFacet.sol
knoll3: Addressing the first part, if we don't store an array of archaeologists (or an array of structs) then how would we know which archaeologists are on the sarcophagus? Since the `sarcophagusArchaeologists` is a mapping there is no way to iterate over it. The best we could do is, given an archaeologist address, tell if the archaeologist is on that sarcophagus.

The same applies for the other addresses: embalmer, recipient, and arweaveArchaeologist. If these are not stored on the Sarcophagus the best we could do is, given an address, tell if that address belongs to the sarcophagus. 

Addressing the second part, storing an array of Archaeologist structs in the Sarcophagus object is possible. However, the pattern I'm seeing in solidity is to not store an array of structs but an array of keys (in this case the archaeologist addresses) and then store a mapping of the structs somewhere else.

Please correct me if I'm wrong on any of this.

contracts/facets/EmbalmerFacet.sol
knoll3: Is there a situation where a `uint256` can be undefined? It should only ever be 0 or a positive number, so I think that no validation would be necessary.

contracts/facets/EmbalmerFacet.sol
knoll3: I made this an enum

contracts/facets/EmbalmerFacet.sol
sethhrbek: We can get this info via events, so I think these can be removed now, or can put in a TODO to remove them later.

contracts/facets/EmbalmerFacet.sol
sethhrbek: Okay cool, enum should work great. 

The only derived state we should need is whether a sarcophagus can be resurrected (`m` archaeologists have successfully unwrapped). But even then, we would want the state represented by the enum, the recipient would just need to know the sarcophagus can be resurrected by querying the sarcophagus.

contracts/facets/EmbalmerFacet.sol
sethhrbek: That all makes sense -- if storing an array of structs is an anti-pattern, then makes sense to not do it here. This can stay as written, thanks for laying that out.

sethhrbek (approved):

PR #9, opened 2022-06-15
7 finalize sarcophagus

knoll3:
# Finalize Sarcophagus

Implements the `finalizeSarcophagus` function in `EmbalmerFacet.sol`. 

### Notes
- The minimum shards required was not added to the intializeSarcophagus function yet. I'm getting a stack too deep error which will require some redesign of the function. I'll work on that next. 

### Testing
- Run the automated tests with `npx hardhat test`


contracts/facets/EmbalmerFacet.sol
```diff
@@ -64,32 +96,51 @@ contract EmbalmerFacet {
                 cursedBondAmount
             );
 
-            // Add to the neccessary data structures
+            // If the archaeologist is the arweave archaeologist, set the
+            // storage fee. This is the only storage fee we care about.
+            if (archaeologists[i].archAddress == arweaveArchaeologist) {
+                storageFee = archaeologists[i].storageFee;
+            }
```
sethhrbek, line 103: Does there need to be any validation that the arweaveArchaeologist address *is* present in one of the archeologist structs? Otherwise this storage fee would never get set.

alternatively, it could validate the storage fee is not 0 outside of this loop, which means this conditional returned true at some point.

contracts/facets/EmbalmerFacet.sol
```diff
@@ -121,4 +170,152 @@ contract EmbalmerFacet {
...
+            s.sarcophaguses[identifier].archaeologists.length - 1
+        ) {
+            revert LibErrors.IncorrectNumberOfArchaeologistSignatures(
+                archaeologistSignatures.length
+            );
+        }
+
+        // Iterate over each regular archaeologist signature. This will not
+        // include the arweave archaeologist.
+        for (uint256 i = 0; i < archaeologistSignatures.length; i++) {
```
sethhrbek, line 244: The only extra validation to consider would be the uniqueness of the signatures. Could theoretically pass in the same signature 4 times here, and this wouldn't revert.

contracts/facets/EmbalmerFacet.sol
```diff
@@ -46,11 +62,27 @@ contract EmbalmerFacet {
...
+        // Initialize the storage fee of the archaeologist who uploades to
+        // arweave. This will be obtained in the for loop and stored on the
+        // sarcophagus object.
+        uint256 storageFee = 0;
+
+        // We need to iterate over every archaeologist to
+        //   - lock up archaeologist's free bond
+        //   - get the storage fee from the arweave archaeologist
+        //   - store each archaeologist's bounty, digging fee, and unencrypted shard in app storage
+        //   - get the archaeologist's address to store on the sarcophagus
```
sethhrbek, line 85: I missed this during previous review -- I recall us discussing locking up the arch's cursed bond amounts during the `finalize` transaction (instead of in `initialize`)

This was b/c once those cursed bonds are locked, they can only be returned if the embalmer cancels the sarcophagus. If this is moved to `finalize`, then the archaeologists can guarantee the sarcophagus is "live" if those cursed bonds are locked.

The alternative would be to allow archaeologists to call `cancel` on a sarcophagus after some time period (maybe 24 hrs), if we wanted to leave the locking up of bonds here.

sethhrbek:
Merging in now to unblock another branch, comments will be handled in another PR (if  necessary)

contracts/facets/EmbalmerFacet.sol
knoll3: Very good point. I think we should go with the latter option to avoid iterating over the list of archaeologists.

contracts/facets/EmbalmerFacet.sol
knoll3: Yes, I remember now. It makes more sense to have this in finalize.

contracts/facets/EmbalmerFacet.sol
knoll3: Yet another astute observation. I'll have to think about the best way to do this.

PR #12, opened 2022-06-15
11 add minshards to initializesarcophagus

knoll3:
# Minshards and sarcoToken modification

This PR is rebased on #9.

### minShards
Adds `minShards` to `initalizeSarcophagus` which is the minimum number of shards required to derive the embalmer's private key. This is the `m` in `m of n`. 

### sarcoToken
Previously the sarcoToken was passed into `initializeSarcophagus` and `finalizeSarcophagus`. Now the sarcoToken is being initialized immediately after the diamond cut for the sarophagus facets. 
In `deploy-diamond.ts`, `calldata` is passed into the diamond cut function. This calls what function is included in calldata with its arguments after a diamond cut happens. 
The first diamond cut in the deploy scripts performs the diamond cut for the facets required for the diamond pattern to work. In this case the calldata is the `init()` function in `DiamondInit.sol`. 
The second diamond cut in the deploy scripts performs the diamond cut for the sarcophagus specific facets (`EmbalmerFacet.so`l and `ArchaeologistFacet.sol`). In this case the calldata is the `init()` function in `AppStorageInit.sol`. The `AppStorageInit.sol` is responsible for initializing any values in AppStorage, like `sarcoToken`. The sarco token may now be accessed from anywhere AppStorage is available. 

According to the diamond pattern, this is exactly what calldata is intended for. There are other ways to initialize the sarco token, but this way keeps the sarcophagus specific code out of the diamond pattern code.

### DIFF
--- a/contracts/facets/EmbalmerFacet.sol
+++ b/contracts/facets/EmbalmerFacet.sol
@@ -36,8 +36,8 @@ contract EmbalmerFacet {
     /// @param arweaveArchaeologist The address of the archaeologist who uploads to arweave
     /// @param recipient the address of the recipient
     /// @param resurrectionTime the resurrection time of the sarcophagus
-    /// @param sarcoToken The erc20 sarcophagus token
     /// @param canBeTransferred Whether the sarcophagus can be transferred
+    /// @param minShards The minimum number of shards
     /// @return The index of the new sarcophagus
     function initializeSarcophagus(
         string memory name,
@@ -46,8 +46,8 @@ contract EmbalmerFacet {
         address arweaveArchaeologist,
         address recipient,
         uint256 resurrectionTime,
-        IERC20 sarcoToken,
-        bool canBeTransferred
+        bool canBeTransferred,
+        uint8 minShards
     ) external returns (uint256) {
         // Confirm that this exact sarcophagus does not already exist
         if (
@@ -65,6 +65,16 @@ contract EmbalmerFacet {
         // Confirm that archaeologists are provided
         if (archaeologists.length == 0) {
             revert LibErrors.NoArchaeologistsProvided();
+        }
+
+        // Confirm that minShards to less than the number of archaeologists
+        if (minShards > archaeologists.length) {
+            revert LibErrors.MinShardsGreaterThanArchaeologists(minShards);
+        }
+
+        // Confirm that minShards is greater than 0
+        if (minShards == 0) {
+            revert LibErrors.MinShardsZero();
         }
 
         // Initialize a list of archaeologist addresses to be passed in to the
@@ -131,6 +141,7 @@ contract EmbalmerFacet {
             name: name,
             state: LibTypes.SarcophagusState.Exists,
             canBeTransferred: canBeTransferred,
+            minShards: minShards,
             resurrectionTime: resurrectionTime,
             arweaveTxId: "",
             storageFee: storageFee,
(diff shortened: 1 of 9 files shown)


contracts/facets/EmbalmerFacet.sol
```diff
@@ -153,7 +164,7 @@ contract EmbalmerFacet {
         );
 
         // Transfer the total fees amount in sarco token from the msg.sender to this contract
-        sarcoToken.transferFrom(msg.sender, address(this), totalFees);
+        s.sarcoToken.transferFrom(msg.sender, address(this), totalFees);
 
         // Emit the event
         emit LibEvents.InitializeSarcophagus(
```
sethhrbek, line 170: Can `minShards` be included in the emitted event?

scripts/deploy-diamond.ts
```diff
@@ -23,9 +23,7 @@ const createAppDiamondCuts = async (): Promise<DiamondCut[]> => {
   // Deploy Embalmer Facet
```
sethhrbek, line 23: Unrelated to this PR -- do we need the private keys library anymore (lines 14-16) since unencrypted shards are used for unwrapping?

scripts/deploy-diamond.ts
```diff
@@ -165,18 +173,40 @@ export const deployDiamond = async () => {
...
+  const diamondInitCallData = diamondInit.interface.encodeFunctionData("init");
+
+  // This adds the sarco token to AppStorage.
+  //
+  // Encode the data for appStorageInit init function call The
+  // AppStorageInit.init() function will be called with delegatecall after the
+  // diamond cut for the sarcophagus facets is performed.
+  const appInitCallData = appStorageInit.interface.encodeFunctionData("init", [
+    mockSarcoToken.address,
+  ]);
```
sethhrbek, line 190: When deploying to testnet / mainnet, would we be running this same script, but passing in the testnet / mainnet sarco token address? Or would a different script be run?

test/facets/embalmer-facet.spec.ts
```diff
@@ -97,8 +98,8 @@ describe("Contract: EmbalmerFacet", () => {
         arweaveArchaeologist.address,
         recipient.address,
         resurrectionTime,
-        sarcoToken.address,
-        canBeTransfered
+        canBeTransfered,
```
sethhrbek, line 101: Misspelling here (minor since this is a test, doesn't really matter).

contracts/facets/EmbalmerFacet.sol
knoll3: Will do.

scripts/deploy-diamond.ts
knoll3: I don't think we need the private keys library anymore

scripts/deploy-diamond.ts
knoll3: This script should be used for deploying to testnet/mainnet. If not it would be very similar.

contracts/facets/EmbalmerFacet.sol
knoll3: Scratch that. Getting the stack too deep error again. I'll have to think of another way.

test/facets/embalmer-facet.spec.ts
knoll3: Fixed.

contracts/facets/EmbalmerFacet.sol
sethhrbek: @knoll3 -- dont worry about making this update --- we can get it off the contract data if we need to

PR #13, opened 2022-06-16
10 cancel bury

knoll3:
# Cancel and Bury
Adds `cancelSarcophagus` and `burySarcophagus` to `EmbalmerFacet.sol`. 

### Notes
- Updated solidity from version `0.8.9` to `0.8.13`.
- Fixed issues brought up in #7.

### DIFF
--- a/contracts/facets/EmbalmerFacet.sol
+++ b/contracts/facets/EmbalmerFacet.sol
@@ -170,8 +170,8 @@ contract EmbalmerFacet {
         // Calculate the total fees in sarco tokens that the contract will
         // receive from the embalmer
         uint256 totalFees = LibBonds.calculateTotalFees(
-            archaeologists,
-            storageFee
+            identifier,
+            archaeologistAddresses
         );
 
         // Transfer the total fees amount in sarco token from the msg.sender to this contract
--- a/contracts/facets/EmbalmerFacet.sol
+++ b/contracts/facets/EmbalmerFacet.sol
@@ -323,7 +323,8 @@ contract EmbalmerFacet {
     }
 
     /// @notice Cancels a sarcophagus. An embalmer may cancel a sarcophagus after
-    /// `initializeSarcophagus` but before `finalizeSarcophagus`.
+    /// `initializeSarcophagus` but before `finalizeSarcophagus`. The embalmer's
+    /// fees that were locked up will be refunded.
     /// @param identifier the identifier of the sarcophagus
     /// @return The boolean true if the operation was successful
 
--- a/contracts/facets/EmbalmerFacet.sol
+++ b/contracts/facets/EmbalmerFacet.sol
@@ -322,6 +322,76 @@ contract EmbalmerFacet {
         return true;
     }
 
+    /// @notice Cancels a sarcophagus. An embalmer may cancel a sarcophagus after
+    /// `initializeSarcophagus` but before `finalizeSarcophagus`.
+    /// @param identifier the identifier of the sarcophagus
+    /// @return The boolean true if the operation was successful
+
+    function cancelSarcophagus(bytes32 identifier) external returns (bool) {
+        // Confirm that the sender is the embalmer
+        if (s.sarcophaguses[identifier].embalmer != msg.sender) {
+            revert LibErrors.SenderNotEmbalmer(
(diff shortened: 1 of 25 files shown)


contracts/facets/EmbalmerFacet.sol
```diff
@@ -120,6 +114,17 @@ contract EmbalmerFacet {
...
+            // checking that the archaeologist does not already exist from
+            // previous iterations in this loop.
+            if (
+                archaeologistExists(identifier, archaeologists[i].archAddress)
+            ) {
+                revert LibErrors.ArchaeologistListNotUnique(
+                    archaeologistAddresses
+                );
+            }
+
```
sethhrbek, line 127: (minor) -- this could possibly be the first logic run in this loop (before state is set)

contracts/facets/EmbalmerFacet.sol
```diff
@@ -136,6 +141,15 @@ contract EmbalmerFacet {
...
         }
 
+        // If the storage fee is 0, then the storage fee was never set since the
+        // default value is 0. This means that either the arweave archaeologist
+        // was not included in the list of archaeologists or the arweave
+        // archaeologist set their storage fee to 0. In either case the
+        // transaction should be reverted.
+        if (storageFee == 0) {
+            revert LibErrors.ArweaveArchaeologistNotInList();
+        }
```
sethhrbek, line 151: The only edge case here would be if an archaeologist set their `FEE_PER_BYTE` value to 0, thus making an arweave TX free for the embalmer (although the arch would still pay the arweave TX fee when they upload the file)

We may need to require this `FEE_PER_BYTE` *not* be 0 for the archaeologist config, or make the front-end still pay some insignificant amount (e.g. 0.000001 SARCO) in the event `FEE_PER_BYTE` is set to 0.

contracts/facets/EmbalmerFacet.sol
```diff
@@ -311,6 +356,126 @@ contract EmbalmerFacet {
...
+
+        // Set the sarcophagus state to done
+        s.sarcophaguses[identifier].state = LibTypes.SarcophagusState.Done;
+
+        address[] memory archaeologistAddresses = s
+            .sarcophaguses[identifier]
+            .archaeologists;
+
+        // Re-calculate the total fees that the embalmer locked up in initializeSarcophagus
+        uint256 storageFees = LibBonds.calculateTotalFees(
```
sethhrbek, line 395: This variable name should probably be `totalFees`, since it includes all fees (not just storage)

contracts/facets/EmbalmerFacet.sol
```diff
@@ -311,6 +356,126 @@ contract EmbalmerFacet {
...
+            // Unlock the archaeologist's cursed bond
+            LibBonds.freeArchaeologist(identifier, archaeologistAddresses[i]);
+
+            // Transfer the digging fees to the archaeologist
+            s.sarcoToken.transfer(
+                archaeologistAddresses[i],
+                getArchaeologist(identifier, archaeologistAddresses[i])
+                    .diggingFee
+            );
+        }
```
sethhrbek, line 471: I realize this is the same functionality as in `v1` -- but what happens to the bounty the embalmer locked up for the sarcophagus? Is that in the contract forever? 

Wondering if we should either transfer that back to the embalmer or potentially make it a protocol fee.

sethhrbek (commented):
Left a few comments -- this is probably good to merge, but we can discuss on standup

contracts/facets/EmbalmerFacet.sol
knoll3: Good catch. Not sure how that happened

contracts/facets/EmbalmerFacet.sol
knoll3: Moved

contracts/facets/EmbalmerFacet.sol
knoll3: I was also thinking that the archaeologist service would validate this number is not 0. We should probably set the default FEE_PER_BYTE value to a realistic number in the arch service.

sethhrbek (approved):

PR #17, opened 2022-06-21
16 unwrap sarcophagus

knoll3:
# Unwrap Sarcophagus

Adds the `unwrapSarcophagus()` function to `ArchaeologistFacet.sol`. 

### Notes
- The tests don't use Shamir Secret Sharing since the contracts don't care at all about the contents of the unencrypted shards. The contract only checks that the shards passed in during initialize match the shards being passed in during unwrap.

### DIFF
--- a/contracts/facets/ArchaeologistFacet.sol
+++ b/contracts/facets/ArchaeologistFacet.sol
@@ -2,6 +2,8 @@
 pragma solidity ^0.8.13;
 
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "../libraries/LibTypes.sol";
+import {LibUtils} from "../libraries/LibUtils.sol";
 import {LibEvents} from "../libraries/LibEvents.sol";
 import {LibErrors} from "../libraries/LibErrors.sol";
 import {LibBonds} from "../libraries/LibBonds.sol";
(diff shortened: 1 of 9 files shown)


contracts/facets/ArchaeologistFacet.sol
```diff
@@ -84,4 +86,80 @@ contract ArchaeologistFacet {
...
+            );
+        }
+
+        // Store the unencrypted shard in on the archaeologist object in the sarcophagus
+        s
+        .sarcophagusArchaeologists[identifier][msg.sender]
+            .unencryptedShard = unencryptedShard;
+
+        // Set the sarcophagus state as done
+        s.sarcophaguses[identifier].state = LibTypes.SarcophagusState.Done;
```
sethhrbek, line 146: This should only be set to done if *all* of the archaeologists have completed their job. I would say this could be removed for now, and we should talk through *if* a successful sarcophagus needs to have its state changed (from "Exists")

contracts/facets/ArchaeologistFacet.sol
```diff
@@ -84,4 +86,80 @@ contract ArchaeologistFacet {
...
+            s.sarcophaguses[identifier].resurrectionWindow
+        );
+
+        // Confirm that the archaeologist has not already unwrapped by checking if the unencryptedShard is empty
+        LibUtils.archaeologistUnwrappedCheck(identifier, msg.sender);
+
+        // Comfirm that the sarcophagus has been finalized
+        if (bytes(s.sarcophaguses[identifier].arweaveTxId).length == 0) {
+            revert LibErrors.SarcophagusNotFinalized(identifier);
+        }
```
sethhrbek, line 126: This could potentially get moved to a library (something like `isFinalized(identifier)`), since this logic is used in multiple places.

sethhrbek: Ignore this comment -- I see you did this in a followup PR

contracts/facets/ArchaeologistFacet.sol
```diff
@@ -84,4 +86,77 @@ contract ArchaeologistFacet {
...
+    /// @notice Unwraps the sarcophagus.
+    /// @dev Verifies that the unencrypted shard matches the hashedShard stored
+    /// on chain and pays the archaeologist.
+    /// @param identifier The identifier of the sarcophagus
+    /// @param unencryptedShard The archaeologist's corresponding unencrypted shard
+    /// @return The boolean true if the operation was successful
+    function unwrapSarcophagus(
+        bytes32 identifier,
+        bytes memory unencryptedShard
+    ) external returns (bool) {
```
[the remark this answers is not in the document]
sethhrbek: You are correct it's probably not necessary. It may be we wanted the caller to have *some* value returned (which would always be true). Once we see how the caller catches errors (or reverts), we can remove this return value.

PR #21, opened 2022-06-21
8 clean accuse

DarksightKellar:
- clean should be complete
- accuse is missing functionality to reimburse unaccused archaeologists
- Removed unneeded params from some ArchaeologistFacet functions (and associated tests)
- For suggestion on getting rid of nested for loop in accuse function, see commit 707f1b9


contracts/facets/ThirdPartyFacet.sol
```diff
@@ -0,0 +1,254 @@
...
+        // Figure out which archaeoligists did not fulfil their duties;
+        // accumulate their digging fees and bounties
+        address[] memory archAddresses = sarco.archaeologists;
+
+        uint256 totalCursedBond;
+        uint256 totalDiggingFee;
+        uint256 totalBounty;
+
+        for (uint256 i = 0; i < archAddresses.length; i++) {
+            if (!s.archaeologistSuccesses[archAddresses[i]][identifier]) {
```
knoll3, line 43: In a PR that hasn't been merged yet I added the unencrypted shards to the `archaeologistStorage` struct. Checking if an archaeologist has uploaded an unencrypted shard is another way to check if an archaeologist has done their job. 

We could create a function that pulls the unencrypted shard like this
```
bytes memory unencryptedShard = s.sarcophagusArchaeologists[identifier][archAddresses[i]].unencryptedShard;
```
And then checks that it's length is greater than 0.

Just leaving this note here so that we can consider changing it after everything else is merged.

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -0,0 +1,254 @@
...
+
+        emit LibEvents.CleanUpSarcophagus(
+            identifier,
+            msg.sender,
+            cleanerBondReward,
+            embalmerBondReward
+        );
+    }
+
+    function accuse(
```
knoll3, line 81: Would you mind documenting this function like the others please?

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -0,0 +1,254 @@
...
+        bytes32 sarcoId,
+        bytes32[] memory unencryptedShards,
```
knoll3, line 83: The type of unencryptedShards should actually be `bytes[]`.

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -0,0 +1,254 @@
...
+            }
+        }
+
+        // Reaching this point means accusal is good to go
+
+        uint256 diggingFeesToBeDistributed = 0;
+        uint256 bountyToBeDistributed = 0;
+        uint256 totalCursedBond = 0;
+
+        for (uint256 i = 0; i < accusedArchs.length; i++) {
```
knoll3, line 131: Could the contents of this for loop be combined with the previous loop instead of having two loops?

contracts/facets/ThirdPartyFacet.sol
DarksightKellar: That seems like a better way to do this, yep. Maybe create a task for after the PR is merged?

test/utils/time.ts
```diff
@@ -0,0 +1,50 @@
+import { ethers } from "hardhat";
```
knoll3, line 1: Creating these functions was a great idea.

test/facets/third-party-facet.spec.ts
```diff
@@ -0,0 +1,332 @@
...
+            paymentDestinationBalance = await sarcoToken.balanceOf(paymentAccount.address);
+
+            // after cleaning...
+            expect(embalmerBalanceAfter.gt(embalmerBalanceBefore)).to.be.true;
+            expect(paymentDestinationBalance.gt(0)).to.be.true;
+
+            //TODO: Actually calculate sum, and verify on exact amounts instead
+        });
+
+        it("Should emit CleanUpSarcophagus on successful cleanup", async () => {
```
knoll3, line 201: I know we talked about this in stand up. These event tests should probably be left out until we find a solution so we don't get failing tests.

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -0,0 +1,254 @@
...
+        // For each provided shard, search through each archaeologist's hashed shard
+        // to see if the provided hard's hash matches one on storage. If so, flag that
+        // archaeologist as accusable
+        uint256 pos = 0;
+        for (uint256 i = 0; i < unencryptedShards.length; i++) {
+            bytes32 shardHash = _hashHelper(unencryptedShards[i]);
+            bool matchingHash = false;
+
+            // Ew
+            for (uint256 j = 0; j < bondedArchsAddresses.length; j++) {
```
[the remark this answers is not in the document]
DarksightKellar: Good to know those are your initial thoughts too.  That was the point of the suggestion to add `hashedShardArchaeologists[shardHash`.

My only concern was if the potential data duplication was worth it (and if the pattern made sense it the first place).

test/utils/time.ts
adamgall: lol thanks for the shout out -- i'm sure i cribbed some version of this from elsewhere on the internet though ;)

we're all standing on the shoulders of giants

contracts/facets/ThirdPartyFacet.sol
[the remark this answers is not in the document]
knoll3: I've noticed that the data duplication you're talking about is sometimes common in solidity to save on transaction costs. We are basically saving a decent amount of memory at the cost of using a little bit of storage. I'd be interested in seeing what the difference is in gas cost.

knoll3 (approved):

knoll3 (approved):

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -0,0 +1,253 @@
...
+
+        if (sarco.resurrectionTime < block.timestamp) {
+            revert LibErrors.SarcophagusIsUnwrappable();
+        }
+
+        if (unencryptedShards.length < sarco.minShards) {
+            revert LibErrors.NotEnoughProof();
+        }
+
+        address[] memory accusedArchAddresses = new address[](
```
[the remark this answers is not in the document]
DarksightKellar: It's not, currently. But might be when implementing reimburse logic. I'll remove it if it turns out the algorithm can work without it.

I should add a todo comment to make that obvious.

PR #25, opened 2022-06-22
23 protocol fees

knoll3:
# Protocol Fees
Adds a basic implementation of the method to collect protocol fees. 

A constant protocol fee defined in an env file is taken from the embalmer on each rewrap. I'm sure this will become more complex pending community input but this PR provides a starting point. 

### Notes
- Add `PROTOCOL_FEE=10` to your `.env` file.
- This value is added to the contract after the diamond cut and stored in AppStorage as `protocolFee`.
- The total amount of protocol fees collected is stored in `totalProtocolFees`.
- `withdrawProtocolFees()` was added to a new contract named `AdminFacet.sol`.
- Only the contract owner may withdraw protocol fees at this time.

### DIFF
--- a/scripts/deploy-diamond.ts
+++ b/scripts/deploy-diamond.ts
@@ -3,7 +3,7 @@ import { ethers } from "hardhat";
 import { Diamond } from "../typechain";
 import { DiamondCut, FacetCutAction } from "../types";
 
-const protocolFee = process.env.PROTOCOL_FEE;
+const protocolFee = process.env.PROTOCOL_FEE || 0;
 
 /**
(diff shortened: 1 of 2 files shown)


sethhrbek:
This looks good -- we may also want to consider an "only owner" `updateProtocolFee` function to update the fee. But lets hold off on deciding this until we discuss implementation with community.

sethhrbek (approved):

knoll3:
> This looks good -- we may also want to consider an "only owner" `updateProtocolFee` function to update the fee. But lets hold off on deciding this until we discuss implementation with community.

I started to do this and removed it. I agree that we should wait on this.

PR #29, opened 2022-06-24
Pay out non naughty archaeologists

DarksightKellar:
I ended up having to use another loop to do this, and that loop contains a nested loop. 

The outer loop iterates through all archs bonded to the sarco; the inner one iterates through just the accused archs, with the possibility of terminating prematurely.

Also, reimbursement is happening right as non-naughty archs are found, to avoid yet another loop if I were to merely track their addresses in another array.

### DIFF
--- a/contracts/facets/ThirdPartyFacet.sol
+++ b/contracts/facets/ThirdPartyFacet.sol
@@ -282,13 +282,6 @@ contract ThirdPartyFacet {
         // transfer the other half of the cursed bond to the transaction caller
         s.sarcoToken.transfer(paymentAddress, halfToSender);
 
-        // This cannot be (easily) done here.
-        // Instead, it's done as defaulters are being aggregated in clean function
-        // LibBonds.decreaseCursedBond(
-        //     sarc.archaeologist,
-        //     sarc.currentCursedBond
-        // );
-
         return (halfToSender, halfToEmbalmer);
     }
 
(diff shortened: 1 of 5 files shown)


adamgall:
Apologies for jumping in here without much context.

But I'm wondering about a thing that @sethhrbek and I may have spoken about in the past.

The idea that these contracts never "push" money/rewards to archaeologists. But rather, just keeps track of them and requires that archaeologists come and "pull" their money out of the system.

That's a pretty common solidity pattern. Benefits include:

- doesn't block payments for users if one of the "push" payments get stuck for any reason (and there definitely might be reasons)
- doesn't require that one user pays all the gas for pushing payments to a bunch of users

knoll3 (approved):

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -189,9 +210,29 @@ contract ThirdPartyFacet {
...
+    function _reimburseArch(bytes32 sarcoId, address arch) private {
+        LibTypes.ArchaeologistStorage storage goodArch = s
+            .sarcophagusArchaeologists[sarcoId][arch];
+
+        uint256 cursedBond = LibBonds.calculateCursedBond(
+            goodArch.diggingFee,
+            goodArch.bounty
+        );
+
+        s.sarcoToken.transfer(arch, cursedBond);
```
sethhrbek, line 227: @DarksightKellar -- this line should be removed. You may be handling this in your current branch, so I will get this merged but wanted to point it out.

PR #35, opened 2022-06-29
Code cleanup

DarksightKellar:
Leaving out DRY-ing tests for now until @knoll3's updates are ready

### DIFF
--- a/contracts/libraries/LibUtils.sol
+++ b/contracts/libraries/LibUtils.sol
@@ -251,17 +251,6 @@ library LibUtils {
         }
     }
 
-    /**
-     * @notice Reverts if msg.sender is not equal to passed-in address
-     * @param account the account to verify is msg.sender
-     */
-    function sarcophagusUpdater(address account) public view {
-        require(
-            account == msg.sender,
-            "sarcophagus cannot be updated by account"
-        );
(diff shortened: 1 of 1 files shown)


contracts/libraries/LibUtils.sol
```diff
@@ -56,14 +56,12 @@ library LibUtils {
         require(bytes(newAssetId).length > 0, "assetId must not have 0 length");
     }
 
-    function archaeologistUnwrappedCheck(
-        bytes32 identifier,
-        address archaeologist
-    ) internal view {
+    function archaeologistUnwrappedCheck(bytes32 sarcoId, address archaeologist)
```
sethhrbek, line 59: @DarksightKellar  --- should this stay with each argument on its own line? Elsewhere in the code multiple arguments are split onto their own lines. 

this is very minor (and not a blocker)

sethhrbek (approved):

contracts/libraries/LibUtils.sol
DarksightKellar: @sethhrbek -- This is a line length thing, so as a line approaches 100 chars (in this setup), auto-formatting breaks it up

PR #50, opened 2022-07-22
DRAFT: Setup example upgrade

DarksightKellar:
Mostly documentation, and a couple scripts. 
readme has full documentaion


knoll3:
We need to talk about upgradeable contracts again before we do anything with this PR.

PR #54, opened 2022-08-29
Register archaeologist

sethhrbek:
## Overview

Adds registering / updating of archaeologist profiles to the contract. This will allow `initializeSarcophagus` and `finalizeSarcophagus` to be combined into a single function in a followup PR.

Currently there are 2 profile values that can be updated:
`minimumDiggingFee`
`maximumRewrapInterval`

`freeBond` can optionally also be passed in if the archaeologist wants to deposit free bond when they register/update. They can also opt to not pass in a free bond and deposit free bond separately.

## TODO
We need to examine whether the `encryptionPublicKey` should be added to the profile or not. The system can function without this update, but it should be discussed further.

### DIFF
--- a/contracts/facets/ArchaeologistFacet.sol
+++ b/contracts/facets/ArchaeologistFacet.sol
@@ -31,6 +31,13 @@ contract ArchaeologistFacet {
         uint256 freeBond
     );
 
+    event UpdateArchaeologist(
+        address indexed archaeologist,
+        uint256 minimumDiggingFee,
+        uint256 maximumRewrapInterval,
+        uint256 freeBond
+    );
+
     event WithdrawFreeBond(
         address indexed archaeologist,
         uint256 withdrawnBond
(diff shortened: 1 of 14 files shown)


contracts/facets/EmbalmerFacet.sol
```diff
@@ -64,7 +64,7 @@ contract EmbalmerFacet {
     function initializeSarcophagus(
```
DarksightKellar, line 64: Param doc for `sarcoArchaeologists` is still `archaeologists`

contracts/facets/ArchaeologistFacet.sol
```diff
@@ -34,22 +48,88 @@ contract ArchaeologistFacet {
...
+                maximumRewrapInterval: maximumRewrapInterval,
+                freeBond: freeBond,
+                cursedBond: 0,
+                rewards: 0
+            });
+
+        // transfer SARCO tokens from the archaeologist to this contract, to be
+        // used as their free bond. can be 0.
+        if (freeBond > 0) {
+            s.sarcoToken.transferFrom(msg.sender, address(this), freeBond);
```
DarksightKellar, line 73: Is there a way we could easily integrate approval into this process? Anytime this is called without previously having approved spending, this will revert - which I think will happen nearly 100% of the time.

Actually as I'm typing this I'm thinking this is something that the arch service will have to remember to do prior to attempting to register with free bond.

contracts/facets/ArchaeologistFacet.sol
```diff
@@ -34,22 +48,88 @@ contract ArchaeologistFacet {
...
+        }
+
+        emit UpdateArchaeologist(
+            msg.sender,
+            existingArch.minimumDiggingFee,
+            existingArch.maximumRewrapInterval,
+            existingArch.freeBond
+        );
+    }
```
DarksightKellar, line 114: Something occurred to me -- could we not combine register and update into one function? Logically they seem the same, with the only real difference being a "create-if-not-exists" idea. 

On the other hand if `updateArchaeologist` will likely diverge significantly in the future then this comment becomes irrelevant

contracts/facets/ArchaeologistFacet.sol
```diff
@@ -34,22 +48,88 @@ contract ArchaeologistFacet {
...
+
     /// @notice Deposits an archaeologist's free bond to the contract.
     /// @param amount The amount to deposit
     function depositFreeBond(uint256 amount) external {
-        // Increase the archaeolgist's free bond in app storage
+        LibUtils.revertIfArchProfileDoesNotExist(msg.sender);
```
DarksightKellar, line 119: Glad you caught this lol

contracts/facets/ArchaeologistFacet.sol
```diff
@@ -34,22 +48,88 @@ contract ArchaeologistFacet {
         uint256 withdrawnReward
     );
 
+    function registerArchaeologist(
```
DarksightKellar, line 51: Missing doc here

DarksightKellar (approved):

contracts/facets/ArchaeologistFacet.sol
sethhrbek: @DarksightKellar -- yeah, that's a great point. When registering/updating + `freeBond` > 0, the arch service should first check to see if their wallet has an approval set on the SARCO token contract. If it doesn't, it should create an approval TX (and wait for it to confirm) prior to calling register/update.

contracts/facets/ArchaeologistFacet.sol
knoll3: Your comment makes sense. Technically there should be no difference. I have no problem leaving the update function here in case it does diverge like you said.

knoll3 (approved):

contracts/facets/ArchaeologistFacet.sol
sethhrbek: talked about this on standup, but noting here -- these functions will stay separate for now.

PR #57, opened 2022-09-01
remove bounty

sethhrbek:
## Overview
Removes concept of the bounty from the system completely.

## TODO
The `calculateCursedBond` method is currently *just* the digging fee now, so is pointless. This method should most likely get removed in a followup PR.

The `storageFee` will be removed in the followup PR which also removed the arweave archaeologist.

### DIFF
--- a/contracts/storage/LibAppStorage.sol
+++ b/contracts/storage/LibAppStorage.sol
@@ -41,9 +41,9 @@ struct AppStorage {
     // Bounty, digging fees, storage fees, and the hashed shards of the
     // archaeologists all need to be stored per sarcophagus. This mapping of a
     // mapping stores the archaeologist's data we need per sarcophagus.
-    // Example usage (to retrieve the bounty an archaeologist may claim on some sarcophagus):
+    // Example usage (to retrieve the digging fees an archaeologist may claim on some sarcophagus):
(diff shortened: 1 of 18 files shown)


test/facets/third-party-facet.spec.ts
```diff
@@ -50,24 +50,23 @@ describe("Contract: ThirdPartyFacet", () => {
...
           .reduce(
-            (acc, arch) => [acc[0].add(arch.diggingFee), acc[1].add(arch.bounty)],
-            [BigNumber.from("0"), BigNumber.from("0")]
+            (acc, arch) => [acc[0].add(arch.diggingFee)],
+            [BigNumber.from("0")]
           );
 
-        const totalDiggingFees = totalDiggingFeesAndBounty[0];
-        const totalBounty = totalDiggingFeesAndBounty[1];
+        const totalDiggingFees = sumDiggingFees[0];
```
sethhrbek, line 60: We may not need both these vars now (`sumDiggingFees` / `totalDiggingFees`) -- will look into combining in a followup PR.

contracts/storage/LibAppStorage.sol
```diff
@@ -41,9 +41,9 @@ struct AppStorage {
     // Bounty, digging fees, storage fees, and the hashed shards of the
```
knoll3, line 41: Remove bounty from comment

knoll3 (approved):

DarksightKellar (approved):

PR #59, opened 2022-09-12
Combine initialize finalize

sethhrbek:
**Overview**
Combines `initializeSarcophagus` and `finalizeSarcophagus` into a single `createSarcophagus` function.

**Other Changes**
1. Removes minting of NFTs and updating of metadata from any functions called by any outside party. The code to create the NFTs still remains -- will need a followup PR to allow the archaeologist to optionally mint NFTs.

2. Updates the single hashed shard passed in for each archaeologist on the contracts during `createSarcophagus` to be the double hashed shard. The single hashed shard can be used to accuse an archaeologist, so we don't want to include this as an arg.

3. Deleted/Updated/Added a few errors as necessary.

4. Refactored the specs for the new `create` workflow, and added specs to verify the archaeologist profile values are enforced during `create`.


contracts/facets/EmbalmerFacet.sol
```diff
@@ -79,68 +75,93 @@ contract EmbalmerFacet {
...
 
-        // Confirm that minShards is greater than 0
-        if (sarcophagus.minShards == 0) {
-            revert LibErrors.MinShardsZero();
-        }
-
         // Initialize a list of archaeologist addresses to be passed in to the
         // sarcophagus object
-        address[] memory archaeologistsToBond = new address[](
+        address[] memory cursedArchaeologists = new address[](
```
sethhrbek, line 102: I renamed this only b/c the `archaeologistsToBond` variable is used **after** the archs have already been bonded (as well as before).

contracts/libraries/LibUtils.sol
```diff
@@ -302,33 +236,53 @@ library LibUtils {
...
+    /// is within the window that an archaeologist will accept
+    ///
+    /// @param resurrectionTime the resurrectionTime supplied for the sarcophagus
+    /// @param archaeologist the archaeologist to check the max rewrap interval of
+    function revertIfResurrectionTimeTooFarInFuture(uint256 resurrectionTime, address archaeologist)
+        internal
+        view
+    {
+        AppStorage storage s = LibAppStorage.getAppStorage();
+        uint256 maxResurrectionTime = block.timestamp + s.archaeologistProfiles[archaeologist].maximumRewrapInterval;
```
DarksightKellar, line 266: Love the logical distinction we can now make between (max) resurrection *time* and (max) *rewrap interval* because of your simple renames 🔥

test/facets/embalmer-facet.spec.ts
```diff
@@ -40,44 +38,105 @@ describe("Contract: EmbalmerFacet", () => {
         );
       });
 
-      it("should emit InitializeSarcophagus()", async () => {
-        const { embalmerFacet, initializeTx } = await createSarcoFixture(
-          { shares, threshold, skipFinalize: true },
+      it("should emit createSarcophagus()", async () => {
```
DarksightKellar, line 41: Minor nitpick but I think conventionally events are PascalCased

test/facets/third-party-facet.spec.ts
```diff
@@ -242,14 +243,14 @@ describe("Contract: ThirdPartyFacet", () => {
 
         const tx = thirdPartyFacet.connect(thirdParty).accuse(
           sarcoId,
-          archaeologists.slice(0, threshold).map(a => a.hashedShard),
+          archaeologists.slice(0, threshold).map(a => hashBytes(a.unencryptedShard)),
```
DarksightKellar, line 246: Sorry if this is a dumb question but is this meant to be an unencrypted shard *hash*, or the plaintext shard?

DarksightKellar (approved):

test/facets/embalmer-facet.spec.ts
sethhrbek: good catch, updating this. What's interesting is that the event name was PascalCased in the contract, but this spec still passed -- the `expect.to.emit` must not check for capitalization.

test/facets/third-party-facet.spec.ts
sethhrbek: @DarksightKellar -- `accuse` takes the shard single hash as proof.

This code that was updated here is functionally the same as before. I had to make the update b/c the `createSarcoFixture` no longer returns the shard single hash -- instead it returns the shard double hash (since that is used during `create` now, and we only need the single hash during accuse).

DarksightKellar: Ahhh, gotcha

PR #61, opened 2022-09-23
Protocol fees and rewards

sethhrbek:
## Overview
1. Removes rewards from archaeologist profile -- continue to use the mapping in storage (as was already the case).

2. Remove `LibRewards` -- write rewards code inline, sine most of it is one-liners. I'm assuming this is slightly cheaper on gas, but I should have done some tests here to confirm. Will followup with this in a gas golfing task later on.

3. Adds protocol fees to `createSarcophagus` and `rewrapSarcophagus`. `protocolFee` is renamed to `protocolFeeBasePercentage` -- this is the percentage used to calculate protocol fees. Defaulted this to `1` in the example env file and the deployment file.


.env.example
```diff
@@ -1,5 +1,5 @@
 ETHERSCAN_API_KEY=ABC123ABC123ABC123ABC123ABC123ABC1
 ROPSTEN_URL=https://eth-ropsten.alchemyapi.io/v2/<YOUR ALCHEMY KEY>
 PRIVATE_KEY=0xabc123abc123abc123abc123abc123abc123abc123abc123abc123abc123abc1
-PROTOCOL_FEE=10
+PROTOCOL_FEE=1
```
DarksightKellar, line 4: Looks like you forgot to rename this env variable

contracts/facets/AdminFacet.sol
```diff
@@ -20,4 +20,11 @@ contract AdminFacet {
         // Transfer the protocol fee amount to the sender after setting state
         s.sarcoToken.transfer(msg.sender, totalProtocolFees);
     }
+
+    /// @notice Sets the protocol fee, expressed as a percentage (i.e. 1 = 1%).
+    /// @dev Can only be called by the owner.
+    function setProtocolFee(uint256 protocolFeeBasePercentage) external {
```
DarksightKellar, line 26: Probably should rename this method too, to match its argument

contracts/storage/LibAppStorage.sol
```diff
@@ -9,9 +9,9 @@ import "../libraries/LibTypes.sol";
 struct AppStorage {
     IERC20 sarcoToken;
     ICurses curses;
-    // The amount to be taken from the embalmer each time a protocol fee should
-    // be collected
-    uint256 protocolFee;
+    // The percentage (i.e. 1 = 1%) of a sarcophagus' total digging fees that will be collected on
+    // createSarcophagus and rewrapSarcophagus, paid by the embalmer
+    uint256 protocolFeeBasePercentage;
```
DarksightKellar, line 14: It just occurred to me -- it might be a good idea to also have this explanation in the natspec of `getProtocolFeeBasePercentage`. Just so peeps browsing the contract have a more intuitive location to find a description

deploy/001_sarcophagus_facets.ts
```diff
@@ -37,6 +37,9 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   // deployed.
   // The only reason for doing diamond.deploy again is so we can execute
   // AppStorageInit. This is pretty much just just a convenience.
+  // Protocol fee defaults to 1% (100bps)
+  const protocolFee = "1";
```
DarksightKellar, line 41: base percentage rename here too

DarksightKellar (approved):
Noice.

Review comments are a couple of missed renames and one documentation suggestion

knoll3 (approved):

sethhrbek:
@DarksightKellar -- comments are addressed -- my `protocolFee` rename task was a rush job, thanks so much for catching those misses.

PR #67, opened 2022-09-27
Add getArchaeologistProfiles function

knoll3:
Adds a function called `getArchaeologistProfiles`. This function gets all archaeologist profiles given a list of archaeologist addresses. Linked to this PR in the web app https://github.com/sarcophagus-org/sarcophagus-v2-app/pull/74.


contracts/facets/ViewStateFacet.sol
```diff
@@ -19,6 +19,31 @@ contract ViewStateFacet {
...
+    /// @param addresses The list of archaeologist addresses
+    /// @return The list of archaeologist profiles
+    function getArchaeologistProfiles(address[] memory addresses)
+        external
+        view
+        returns (LibTypes.ArchaeologistProfile[] memory)
+    {
+        LibTypes.ArchaeologistProfile[]
+            memory profiles = new LibTypes.ArchaeologistProfile[](
+                s.archaeologistProfileAddresses.length
```
DarksightKellar, line 33: This here brought a more general question to mind:

Do you think it's possible to abuse contract functions that accept arrays (and especially use the size of that array to create _another_ array like in this case) by calling it with an absurdly sized array of garbage data? How will that affect the nodes on the network? And if it won't, how is the EVM dealing with it?

Or maybe it's just not an "attack" that's worth it in the first place? Like, to what end exactly lol

cc @sethhrbek @adamgall

DarksightKellar (approved):

contracts/facets/ViewStateFacet.sol
sethhrbek: @DarksightKellar -- see: https://ethereum.stackexchange.com/questions/9824/can-solidity-view-pure-functions-be-arbitrarily-complex

> view/pure functions are limited by the gas provided to it. They still "use" gas, even though the sender (from account) doesn't get "charged" for the gas.

So there will be a limit to how large the array is that can be returned. This is the reason this function takes an `addresses[]` array instead of just returning the entirety of the archaeologist profiles, it will let us batch view state calls to a specific size return length.

sethhrbek: To answer your question about abuse -- this view computation would be performed on the own callers node (in our case infura), so we would only be abusing ourselves, not the ethereum network.

sethhrbek (approved):

contracts/facets/ViewStateFacet.sol
DarksightKellar: Ah I see. This was very helpful, thanks!

knoll3: Great question to bring up. Thanks for looking into that, Seth.

PR #69, opened 2022-09-28
Add function to get arch stats

knoll3:
Adds a function to get archaeologist reputation stats.

### Notes
- Added `archaeologistSuccesses` that keeps a list of sarcoIds like the cancels and accusals. Otherwise there is no easy way to get a list of successes.
- Added an aggregate stats view state function to return all stats in a single function call. This will be the function the front-end uses to grab the stats.
- Updated the view state facet stats functions to return the length of the mappings + added index lookup functions so that sarcoIds for each arch stat category can be queried. These individual stats methods (index / lengths) are not used by the front end currently, but think they hold enough utility/value to add.

### DIFF
--- a/contracts/facets/ViewStateFacet.sol
+++ b/contracts/facets/ViewStateFacet.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.13;
 
 import "../libraries/LibTypes.sol";
+import "hardhat/console.sol";
 import {AppStorage} from "../storage/LibAppStorage.sol";
 
 contract ViewStateFacet {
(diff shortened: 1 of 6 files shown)


DarksightKellar (approved):

contracts/facets/ViewStateFacet.sol
```diff
@@ -44,6 +44,33 @@ contract ViewStateFacet {
...
+            );
+
+        for (uint256 i = 0; i < addresses.length; i++) {
+            statsList[i] = LibTypes.ArchaeologistStatistics(
+                addresses[i],
+                s.archaeologistSuccesses[addresses[i]],
+                s.archaeologistCancels[addresses[i]],
+                s.archaeologistAccusals[addresses[i]]
+            );
+        }
```
sethhrbek, line 69: note -- see my comments on discord -- I think we just need the lengths of the mappings for these, and not the sarcoIds. Also think the original successes mapping can be removed (and any view state functions associated with it).

Issue #65, opened 2022-09-27
Max Rewrap Interval -- Lock this in?


sethhrbek:
**Tasks**
1. Add a `maxRewrapInterval` field to the sarcophagus to `LibTypes.Sarcophagus`
2. Add `maxRewrapInterval` as an field on the `SelectedArchaeologistData` struct that is passed into the `createSarcophagus` method..
3. In `createSarcophagus`, pass in the `maxRewrapInterval` to the `LibUtils.verifyArchaeologistSignature`
4. In `verifyArchaeologistSignature`, add the `maxRewrapInterval` to the data of sig being verified.
5. In `rewrapSarcophagus`, validate that the `block.timestamp - resurrectionTime <= sarcophagus.maxRewrapInterval`, and revert if not -- this ensures new resurrection time provided to rewrap is valid.
6. Update/add specs where necessary

Note:
Updating the contracts with these changes will require updates to the archaeologist service and the webapp for the create-sarcophagus workflow.

---
Original notes:

Need to make final decision about how max rewrap interval works.

We had initially decided that sarcophagus would have the maxRewrapInterval *locked* for the duration of the sarcophagus. This value would be derived from the archaeologist with the *smallest* max rewrap interval for the sarcophagus. This max rewrap interval would then be enforced on rewrap.

However --- we have since added archaeologist profiles since this decision, and we could rely on the archaeologists profile values to enforce this. The potential issue here (as it was with v1) is that the archaeologists could arbitrarily change that value and make it impossible for the embalmer to rewrap with the same set of archaeologists.

The further wrinkle is that we *may* be adding a rewrap workflow allowing embalmers to change the set of archaeologists for a sarcophagus, so need to factor in if that affects this decision at all.

PR #71, opened 2022-09-29
65 max rewrap interval

katien:
https://github.com/sarcophagus-org/sarcophagus-v2-contracts/issues/65
- Implements a per sarcophagus maximum resurrection time that will persist for the lifetime of the sarcophagus
- Contracts now require a maximum resurrection time to be signed by each archeologist as part of their curse agreement handshake. Expected format of data to be signed: `[arweaveTxId, unencryptedShardDoubleHash, agreedMaximumRewrapInterval]`
- removes `gracePeriod` from the sarcophagus struct, the application should rely on the system value in `AppStorage` which can be retrieved through the ViewStateFacet

### DIFF
--- a/contracts/facets/EmbalmerFacet.sol
+++ b/contracts/facets/EmbalmerFacet.sol
@@ -41,23 +41,24 @@ contract EmbalmerFacet {
 
     /// @notice Embalmer creates the sarcophagus.
     ///
-    /// The purpose of initializeSarcophagus is to:
+    /// The purpose of createSarcophagus is to:
     ///   - Lock up payment for the selected archaeologists (digging fees)
     ///   - Store the arweave TX IDs pertaining to the encrypted file payload
     ///   -    and the encrypted shards
-    ///   - Verify the selected archaeologists have signed off on their
-    ///   -    unencrypted double hash and the arweave TX ID pertaining
-    ///   -    to the encrypted hashes
+    ///   - Verify the selected archaeologists have signed off on the
+    ///         double hash of their key share,
+    ///         arweave tx id storing key shares,
+    ///         and maximumRewrapInterval to be used for lifetime of the sarcophagus
     ///   - Store the selected archaeologists' addresses, digging fees and
-    ///   -    unencrypted double hashes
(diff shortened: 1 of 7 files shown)


contracts/facets/EmbalmerFacet.sol
```diff
@@ -126,15 +133,15 @@ contract EmbalmerFacet {
 
             // Validate archaeologist profile value requirements are met
             LibUtils.revertIfDiggingFeeTooLow(arch.diggingFee, arch.archAddress);
-            LibUtils.revertIfResurrectionTimeTooFarInFuture(sarcophagus.resurrectionTime, arch.archAddress);
```
sethhrbek, line 129: realizing that we could have left this in and omitted the signature of the agreedMaximumRewrapInterval, but hunch is that this new way probably saves gas

sethhrbek: could also make argument we could now remove `maxRewrapInterval` from the archaeologist profile and have the arch service provide this data to the embalmer web app, since the signature now acts as the on-chain validation of this value.

contracts/facets/EmbalmerFacet.sol
```diff
@@ -126,15 +133,15 @@ contract EmbalmerFacet {
...
 
             totalDiggingFees += arch.diggingFee;
 
-            // Validate the archaeologist has signed off on their shard data:
-            // hashed shard, arweaveTxId[1] (encrypted shard on arweave)
+            // Validate the archaeologist has signed off on the sarcophagus parameters: double hashed key share,
+            // arweaveTxId[1] (tx storing share on arweave), maximumRewrapInterval for sarcophagus
```
sethhrbek, line 140: thanks for updating all these comments

sethhrbek (approved):
this looks great! thanks for refining & fixing the comments.

contracts/facets/EmbalmerFacet.sol
katien: https://github.com/sarcophagus-org/sarcophagus-v2-contracts/issues/73

PR #84, opened 2022-10-10
82 publish abis and interfaces as a node module

katien:


tsconfig.json
```diff
@@ -5,20 +5,7 @@
...
     "outDir": "dist",
-    "declaration": true,
-    "skipLibCheck": true,
-    "forceConsistentCasingInFileNames": true,
+    "declaration": true
   },
-  "allowJs": false,
-  "include": [
-    "./scripts",
-    "./test",
```
DarksightKellar, line 15: Was there a reason these config params were removed (ie `skipLibCheck`, `forceConsistentCasingInFileNames`, `"allowJs": false`, and `./test` in `include`)?

Or do they actually have no real use here, what with javascript type configuration being rubbish and all.

DarksightKellar (commented):
Had a question on `tsconfig` changes, but other than that looks good to me!

tsconfig.json
knoll3: I can't remember why these were added in here to be honest... Some of this looks like some old leftover stuff or possibly even junk copied over from fractal (haven't verified that).

knoll3 (approved):
Should be good after running eslint on some of these files.

tsconfig.json
katien: These configurations were largely taken from fractal with a couple of minor changes to fit our project structure

sethhrbek (approved):

PR #87, opened 2022-10-15
Add transfer to local embalmer script

sethhrbek:
add basic butt script to transfer SARCO locally to embalmer


scripts/transfer.js
```diff
@@ -0,0 +1,18 @@
...
+const hre = require("hardhat");
+
+(async () => {
+  const sarcoToken = await hre.ethers.getContractAt(
+    "SarcoTokenMock",
+    "0x5FbDB2315678afecb367f032d93F642f64180aa3"
+  );
+  const unsignedAccounts = await hre.getUnnamedAccounts();
+  await sarcoToken.transfer(
+    unsignedAccounts[0],
```
DarksightKellar, line 12: Making this an address read from `.env` would transform this script into a one-tap, transfer-to-_anyone_ superpower.

(Also there's a linter warning at line 11 above)

sethhrbek: @DarksightKellar -- good call! I added a transfer address env var and ran the linter.

PR #88, opened 2022-10-18
Some updates to transfer script

DarksightKellar:


scripts/transfer.js
```diff
@@ -1,16 +1,27 @@
...
+// transferAddress defaults to the first unnamed hardhat local node account
 const amountToTransfer = "10000000000000000000000";
-const transferAddress =
-  process.env.TRANSFER_ADDRESS || "0x5FbDB2315678afecb367f032d93F642f64180aa3";
+const mockTokenAddress = "0x5fbdb2315678afecb367f032d93f642f64180aa3";
 
 const hre = require("hardhat");
 
+// Add supported networks to this list on a need basis
+const allowedNetworks = ["localhost"];
```
DarksightKellar, line 9: I was thinking we could add to this list later on whenever we get around to testing on other networks, but to avoid accidentally running on mainnet, or any unexpected network even during tests really. 

Even if the script runner in `package.json` uses `localhost`, this can be run directly

scripts/transfer.js
```diff
@@ -1,16 +1,27 @@
-// Transfer 10,000 sarco tokens to the an address
-// transferAddress defaults to the embalmer address the webapp uses
+// Transfer 10,000 sarco tokens to an address
+// transferAddress defaults to the first unnamed hardhat local node account
 const amountToTransfer = "10000000000000000000000";
-const transferAddress =
-  process.env.TRANSFER_ADDRESS || "0x5FbDB2315678afecb367f032d93F642f64180aa3";
+const mockTokenAddress = "0x5fbdb2315678afecb367f032d93f642f64180aa3";
```
DarksightKellar, line 4: `0x5fbdb2315678afecb367f032d93f642f64180aa3` is actually the mock sarco token address on localhost. Didn't notice until I tried to run this script myself

sethhrbek: @DarksightKellar -- good catch -- note to self -- run the code after making changes.

sethhrbek (approved):
thanks for making these changes @DarksightKellar, they are great

PR #90, opened 2022-10-19
remove all code related to NFTs

katien:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sarcophagus-org/sarcophagus-v2-contracts",
-  "version": "0.1.0",
+  "version": "0.2.0",
```
katien, line 3: I nearly forgot to do this but we'll have to be cautious about it moving forward - breaking changes to the contracts require a version update and publishing the latest version of the package

DarksightKellar: Ah, absolutely.

On the other hand, consumers will not have to also update their dependency, would they, because of the `^`?

sethhrbek (approved):

PR #97, opened 2022-11-09
Update sarco states

DarksightKellar:
This PR updates SarcophagusState to match what will be used in the webapp.

Tests and unwrap functions were also updated to account for this change.


contracts/libraries/LibUtils.sol
```diff
@@ -43,6 +43,23 @@ library LibUtils {
         }
     }
 
+    /**
+     * @notice Reverts if the Sarcophagus does not exist.
+     * @param sarcoId Identifier of the Sarcophagus
+     * @dev A Sarcophagus "exists" if its state is `Active` OR `Resurrecting`. Any other state is interpreted as not existing.
+     */
+    function revertIfNotExists(bytes32 sarcoId) internal view {
```
katien, line 51: Maybe we want to name this something that indicates it's checking for existence of a sarcophagus

DarksightKellar: Absolutely. Done.

katien: actually looking at this more deeply we're going to want to differentiate between nonexistent and what this is checking for (since nonexistent is really just that very first default state in the enum). It sounds like what we're really after is knowing whether or not the sarcophagus is resurrectable, so what do you think of `revertIfNotResurrectable`?

DarksightKellar: I would say let's do that differentiation only if necessary -- as it stands right now, "not exists" really just means "inactive but not one of the `done` states".

I could remove the `Resurrecting` state check, as that was leftover from when I previously had unwrap set state to `Resurrecting` or `Resurrected`, but `Resurrecting` was actually an active state. Actually I might just go ahead and remove that extra check -- `Resurrecting` is now a dynamically returned state that's never stored.

In any case `revertIfSarcoNotExists` is used for more than just in the context of resurrection, so it does fit.

katien: My reasoning is that the sarcophagus object exists on the contract as stored data unless the enum is actually that `0` value of `DoesNotExist` and that terminology will become confusing. 

If we keep it this way, `DoesNotExist` as a state will mean that the sarco with the given ID was never created or stored at all, but NotExists, as determined by this function, will mean that the sarco with the given ID was never created at all OR the sarco was created and then set to Resurrected, Buried, Cleaned, Accused, or Failed.

DarksightKellar: Maybe `Inactive` would be a better term for that 0-th, default state then?

I'm pushing back on attempting to differentiate them at this point since, practically, we don't really care about the difference, nor do we care about that default state on an existing sarcophagus.

If we start to want to do stuff with the `done`-type states, it still wouldn't affect this function, which is used to prevent the contracts from interacting with `done`-ish sarcophagi. 

I'll welcome a name update on the default state, or the function name, or both, but not in the name of differentiating "actually does not exist" from "exists but cannot be used"

DarksightKellar: I think we're saying more or less the same thing lool

katien: No worries, discussion is encouraged 😄 

The 0th state isn't `Inactive` because the sarcophagus never existed, solidity just happens to return an empty object if you request an object from a mapping at a key that hasn't been set. If I call s.sarcophagi["amadeupandnonexitentId"] I'll get back a sarcophagus with state == `DoesNotExist` because that's the 0th state on the enum.

This differentiation will matter because we often want to display sarcophagi that have been created but are in a non-active state. We also allow interactions with the sarcophagi in some of those states. 
 
Example: In the future we're probably going to be building a sarcophagus search page where the user enters a sarcoId and we retrieve the corresponding sarcophagus.
- If the contract call at `ViewStateFacet.getSarcophagus(sarcoId)` returns a sarcophagus object where state == `DoesNotExist`, we don't want to display anything. The ID is invalid and the sarcophagus never existed. 
- If the call returns a sarcophagus object with a done-type state, we'll want to show that sarcophagus to allow viewing and interactions like cleaning if not every archaeologist has supplied their keyshare or been unbonded.

DarksightKellar: Ahh, I see what you mean.

And yet, after giving it some more thought, my mind is circling back to doing simple rename(s) that will be less confusing. I think the logic is sound, also after taking a quick look at how this revert function is being used, currently (note this new `revertIfSarcoNotExists` function simply refactored repeated code blocks).

- `getSarcophagus` does not use `revertIfSarcoNotExists`, so all *actually existing but done* sarcophagi should be returned and viewable by consumers:
```
if (sarco.state != LibTypes.SarcophagusState.Active) {
    return sarco;
}
```

- The `Buried`, `Cleaned` and `Accused` done-type states, which will be set on-chain if any of those are called on a sarco, would and should put those sarco in an eternally not interactable state. In which case `revertIfSarcoNotExists` will work as expected wherever it's being used, as it reverts if the sarco is not `Active`.

- The `Resurrecting`, `Resurrected` and `Failed` states are never actually set on-chain -- they're only dynamically returned from within `getSarcophagus`, and so as far as the contracts are concerned, those sarcophagi that will be returned by `getSarcophagus` as having any of those states are, in fact, `Active`.

...

Aaand now I see where the differentiation will be necessary. `getSarcophagus` *should* revert if the state is the default `DoesNotExist`. 
And I'd say renaming `revertIfSarcoNotExists` to `revertIfSarcoInactive`, *and* a `SarcophagusInactive` error type separate from `SarcophagusDoesNotExist` should address all concerns?

So:
- `revertIfSarcoInactive` will revert with `SarcophagusInactive` if sarco state is not `Active`.

- `getSarcophagus` will revert with `SarcophagusDoesNotExist` if sarco state is `DoesNotExist`, and only there will that happen -- which explicitly means that the sarcophagus actually does not exist in any shape or form.

DarksightKellar: Okay even if usages of `revertIfSarcoInactive` will revert appropriately if the sarco doesn't exist, the error will be `SarcophagusInactive`. I'll add in an extra check to the function so it'll revert with `SarcophagusDoesNotExist` if state is `DoesNotExist `, and continue on if not.

The function will now be known as ... `revertIfNotExistOrInactive`.

Geez I dunno lol.

sethhrbek (approved):

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -29,11 +29,9 @@ contract ThirdPartyFacet {
     /// @param sarcoId The identifier of the sarcophagus to clean
     /// @param paymentAddress The address to which rewards will be sent
     function clean(bytes32 sarcoId, address paymentAddress) external {
-        LibTypes.Sarcophagus storage sarco = s.sarcophagi[sarcoId];
```
katien, line 32: Shoot, only just caught this while working on my own branch but we shouldn't be reverting on inactive sarcophagi here. I'll update.

sethhrbek: @katien -- doesn't the sarcophagus need to be active to clean it? Inactive states would be `Buried`, `Cleaned` or `Accused` -- and don't think any of those states could be cleaned.

contracts/facets/ViewStateFacet.sol
```diff
@@ -249,7 +249,36 @@ contract ViewStateFacet {
         view
         returns (LibTypes.Sarcophagus memory)
     {
-        return s.sarcophagi[sarcoId];
+        LibTypes.Sarcophagus memory sarco = s.sarcophagi[sarcoId];
+        if (sarco.state != LibTypes.SarcophagusState.Active) {
+            return sarco;
+        }
```
sethhrbek, line 255: @katien -- there is an edge case here that isn't handled --

A Sarcophagus could be both `Cleaned` and `Resurrected`, or `Cleaned` and `Failed`. Not sure the best way to handle this -- it's more important the front end shows the `Resurrected`/`Failed` states, but would need to gate the `Clean` function from showing if the sarco has also been cleaned.

PR #101, opened 2022-11-18
96 update accusal flow

katien:


test/facets/third-party-facet.spec.ts
```diff
@@ -1,568 +0,0 @@
-import "@nomiclabs/hardhat-waffle";
```
katien, line 1: Deleted the entire file because we didn't have any other third party facet functions being tested here

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -91,117 +89,124 @@ contract ThirdPartyFacet {
...
+        // be returned to the remaining well behaved archaeologists
+        if (isSarcophagusCompromised) {
+            // update the sarcophagus state to Accused, the outer key is compromised
+            sarco.state = LibTypes.SarcophagusState.Accused;
+
+            // iterate through all archaeologist addresses on the sarcophagus
+            for (uint256 i = 0; i < archaeologistAddresses.length; i++) {
+                // if the archaeologist has never been accused, release their locked bond back to them
+                if (!s.sarcophagusArchaeologists[sarcoId][archaeologistAddresses[i]].accused) {
+                     LibBonds.freeArchaeologist(sarcoId, archaeologistAddresses[i]);
```
katien, line 193: In the original version of this function, we don't refund the digging fees allocated to well-behaved archaeologists back to the embalmer. In this PR I've preserved that functionality but I think we ought to provide the full refund because the archaeologists aren't receiving those digging fees if the sarcophagus is being closed down and the embalmer isn't receiving the services they've paid for.

katien:
Besides the accuse function itself, the bulk of this PR is in accuse.spec.ts which is collapsed by default and comes up at the very end of the other tests so don't miss it! There's some duplicate functionality between the utilities I've created and the existing fixtures but my goal with these utilities is to break things up into smaller pieces so we can more easily generate single components of our tests. They're a WIP and I'll be adding to them as I knock out some of the other contract tasks.

test/facets/thirdparty/accuse.spec.ts
```diff
@@ -0,0 +1,716 @@
...
+            await viewStateFacet.getArchaeologistAccusalsCount(
+              innocentArchaeologist.archAddress
+            );
+          expect(innocentArchaeologistAccusalsCount.toString()).to.equal("0");
+        })
+      );
+    });
+  });
+
+  // todo: can an async context be used with mocha?
```
katien, line 657: I don't think that this is something that can actually be done with mocha, but if anybody has suggestions on better ways to organize a set of tests where there's some asynchronous configuration and then several tests that use that configuration I'd like to hear them. Currently, there's a lot of duplication across tests because I'm having to set up the sarcophagus and accusal call over and over again.

Below is an example of what I'd like to be able to write, but `context` is an alias for `describe` which cannot take an async callback. We have the option to run mocha with `--delay` to do some asynchronous configuration before running the suite but I don't think it gives us the organizational advantage that context/describe do.

contracts/facets/ThirdPartyFacet.sol
DarksightKellar: Agreed

test/facets/thirdparty/accuse.spec.ts
```diff
@@ -0,0 +1,716 @@
...
+      const tx = (await getContracts()).thirdPartyFacet
+        .connect(embalmer)
+        .accuse(sarcophagus.sarcoId, [], embalmer.address);
+
+      await expect(tx).to.be.revertedWith(`SarcophagusIsUnwrappable()`);
+    });
+  });
+
+  describe("Supports accusal of fewer than k archaeologists", function () {
+    it("On a successful accusal of an archaeologist, should transfer the correct amount of funds to embalmer and accuser, slash the archaeologist's bond, mark the arch as accused, and emit an AccuseArchaeologist event", async function () {
```
DarksightKellar, line 63: I'd very much rather split up these (and other similarly combined) assertions into their own `it`-test blocks -- even if it means duplicate code.

Specifically, either a `describe` or `context` with "On successful accusal of an archaeologist", within which we have a distinct `it "should do xyz"` for each expectation.

I'm sure `beforeEach` would be enough to do the prerequisite setup being done below.

It's much easier to digest that way, and easier to debug as well.

test/facets/helpers/accounts.ts
```diff
@@ -0,0 +1,26 @@
...
+  // track unused account index
+  let index = 0;
+  return async (): Promise<SignerWithAddress> => {
+    const unnamedAccounts = await getUnnamedAccounts();
+    if (!unnamedAccounts[index])
+      throw Error(
+        "Not enough test accounts. Increase value in hardhat config."
+      );
+    return await ethers.getSigner(unnamedAccounts[index++]);
+  };
```
DarksightKellar, line 19: Ah, nice

DarksightKellar (changes requested):
Whoo. This is a beefy one, and I'm ashamed to say I didn't read through every line.
The accuse tests are nice and specific, as they should be.

I did start to get the feeling that a number of the new helper functions are already implemented in some form in the existing fixtures. Is the intention to phase them out eventually, or to use both (maybe the new functions solve a different problem)?

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -91,117 +89,124 @@ contract ThirdPartyFacet {
...
         (
             uint256 accuserBondReward,
             uint256 embalmerBondReward
         ) = _distributeLoot(
                 paymentAddress,
                 sarco,
-                totalCursedBond,
-                diggingFeesToBeDistributed
+                totalDiggingFees,
+                totalDiggingFees
```
DarksightKellar, line 207: Could we not now remove one of these params?

test/facets/thirdparty/accuse.spec.ts
DarksightKellar: `beforeEach`, maybe?

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -91,117 +89,124 @@ contract ThirdPartyFacet {
...
+
+
+        address[] memory archaeologistAddresses = s.sarcophagi[sarcoId].archaeologists;
+
+        // the sarcophagus is compromised if k or more archaeologists have been accused successfully on this call
+        bool isSarcophagusCompromised = accusalCount >= sarco.minShards;
+
+        // if the current call hasn't resulted in at least k archaeologists being accused
+        // check if total number of historical accusals on sarcophagus is greater than k
+        uint historicalAccusals = 0;
```
knoll3, line 170: Could we store `historicalAccusals` in sarcoState to avoid the loop below?

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -91,117 +89,124 @@ contract ThirdPartyFacet {
...
+     * and bonds for all remaining archaeologists will be freed
+     *
      * @param sarcoId The identifier of the sarcophagus having leaked shards
-     * @param unencryptedShardHashes At least 'm' unencrypted shard hashes as proof of bad behaviour
+     * @param unencryptedShardHashes hashes of the leaked keyshares
      * @param paymentAddress the address to which rewards should be sent if successful
      */
     function accuse(
         bytes32 sarcoId,
         bytes32[] memory unencryptedShardHashes,
```
knoll3, line 105: What is the purpose for being able to accuse multiple archs in one function? Is it just to save on transaction costs if an accuser has multiple accusals to make?

contracts/facets/ThirdPartyFacet.sol
katien: The first loop only runs through archaeologists that have leaked keyshares and not the full list of archaeologists on the sarcophagus, so unfortunately there's no way to track historical accusations at that stage. Good thought though, I optimized as much as possible but the looping in this function is pretty aggressive anyway.

test/facets/thirdparty/accuse.spec.ts
katien: 100%, I would love to be able to set up this test set with a context and then have an individual `it()` for most of these `expect` calls. I thought about breaking these up into more `it`s and duplicating the setup, but even with utility methods it ends up being a pretty significant amount of boilerplate getting copied between cases.

contracts/facets/ThirdPartyFacet.sol
katien: haha definitely

contracts/facets/ThirdPartyFacet.sol
katien: savings on transaction cost and also if we required accusals to be one at a time, we'd be limiting rewards to the bonds for k archaeologists. With the array, an accuser could get a payout for all n if they'd all leaked their keyshare but with individual calls, the remaining n-k would be freed as soon as the sarcophagus was deemed compromised.

contracts/facets/ThirdPartyFacet.sol
katien: I'm thinking that it's better to have the accuser incur the cost of the loop than it is to add to the data model because accuse isn't going to be called often (and the caller is getting a payout) but an increase in the size of the Sarcophagus model impacts all users.

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -91,117 +89,124 @@ contract ThirdPartyFacet {
...
+            for (uint256 i = 0; i < archaeologistAddresses.length; i++) {
+                if (s.sarcophagusArchaeologists[sarcoId][archaeologistAddresses[i]].accused) {
+                    historicalAccusals++;
                 }
             }
+            // the sarcophagus is compromised if k or more archaeologists have been accused over the lifetime of the sarcophagus
+            if (historicalAccusals >= sarco.minShards) {
+                isSarcophagusCompromised = true;
+            }
+        }
```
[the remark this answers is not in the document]
katien: I don't think we'd actually be able to remove any of the logic other than the part that says to loop over the set of supplied keyshare hashes. We'd still need to take the same actions on the single supplied share, so the logic inside of the loop would stay, and then we'd also need to evaluate whether or not the single archaeologist being accused brought the sarcophagus over the k threshold and if yes free the other archaeologists.

sethhrbek:
@katien -- one more thought --- *if* we were optimizing for gas --

If the accuser passed in an array of archaeologist addresses along with the keyshares, would that help simplify some of the lookups we need to do on the archaeologists?

sethhrbek (approved):

test/facets/thirdparty/accuse.spec.ts
knoll3: I think this is what Kelvin is suggesting 
```
let sarcohpagus: Sarcophagus;
...

beforeEach(async () => {
  sarcohpagus = await getSarcophagus();
  ...
})

it("should do stuff", () => {
  // use sarcophagus var
})
```

knoll3: What you're describing has always been a pain point for me that has made me rip what little hair I have left out of my head. I still haven't come to a satisfactory answer for this, but extracting common code into fixtures is the way I usually go. This is what the current test fixtures are supposed to do, but I'm not sure if they've retained their original purpose over time.

PR #102, opened 2022-11-28
99 getsarcophagus in viewstatefacet

katien:


contracts/facets/EmbalmerFacet.sol
```diff
@@ -34,267 +34,237 @@ contract EmbalmerFacet {
...
-            // Get the archaeolgist's fee data
-            LibTypes.ArchaeologistStorage memory archaeologistData = LibUtils
-                .getArchaeologist(sarcoId, bondedArchaeologists[i]);
-
-            // Transfer the archaeologist's digging fee allocation to the archaeologist's reward pool
-            s.archaeologistRewards[bondedArchaeologists[i]] += archaeologistData.diggingFee;
+        // pay digging fee to each cursed archaeologist on the sarcophagus
+        address[] storage archaeologistAddresses = sarcophagus.cursedArchaeologistAddresses;
+        for (uint256 i = 0; i < archaeologistAddresses.length; i++) {
+            LibTypes.CursedArchaeologist storage cursedArchaeologist = sarcophagus
```
katien, line 261: this needs to be checking that the archaeologist hasn't been accused, handled in future PR

contracts/facets/EmbalmerFacet.sol
```diff
@@ -34,267 +34,237 @@ contract EmbalmerFacet {
...
-            // Add the archaeologist's digging fee to the sum
-            totalDiggingFees += archaeologistData.diggingFee;
-
-            // Update the archaeologist's data in storage
-            s.sarcophagusArchaeologists[sarcoId][
-                bondedArchaeologists[i]
-            ] = archaeologistData;
+            // transfer digging fee to archaeologist's reward pool
+            // todo: consider adding this amount to archaeologistProfile.freeBond instead
+            s.archaeologistRewards[archaeologistAddresses[i]] += cursedArchaeologist.diggingFee;
```
katien, line 266: Consider combining `archaeologistProfile.freeBond` and `s.archaeologistRewards` so that archaeologists can minimize their costs when withdrawing funds and we aren't keeping unnecessary data structures

contracts/facets/ArchaeologistFacet.sol
```diff
@@ -163,58 +155,78 @@ contract ArchaeologistFacet {
...
-
-        // Confirm that the archaeologist has not already unwrapped by checking
-        // if the unencryptedShard is empty
-        LibUtils.archaeologistUnwrappedCheck(sarcoId, msg.sender);
-
-        // Confirm that the sender is an archaeologist on this sarcophagus
-        if (!LibUtils.archaeologistExistsOnSarc(sarcoId, msg.sender)) {
-            revert LibErrors.ArchaeologistNotOnSarcophagus(msg.sender);
+    /// @param rawKeyShare The keyshare the archaeologist is publishing
+    function publishKeyShare(bytes32 sarcoId, bytes calldata rawKeyShare) external {
```
katien, line 164: `calldata` instead of `memory` on reference type parameters saves us a copy operation when the function executes because the arguments aren't having to be copied into memory

.eslintrc.js
```diff
@@ -16,25 +16,14 @@ module.exports = {
...
+    "node/no-unsupported-features/es-syntax": [
+      "error",
+      { ignores: ["modules"] },
+    ],
     "node/no-unpublished-require": 0,
     "node/no-unpublished-import": 0,
     "comma-dangle": "off",
     "@typescript-eslint/indent": "off",
-    // prettier config
-    "prettier/prettier": [
```
katien, line 25: this prettier config was conflicting with the one in .prettierrc

contracts/facets/ArchaeologistFacet.sol
```diff
@@ -11,7 +11,7 @@ import {AppStorage} from "../storage/LibAppStorage.sol";
 contract ArchaeologistFacet {
     AppStorage internal s;
 
-    event UnwrapSarcophagus(bytes32 indexed sarcoId, bytes unencryptedShard);
+    event PublishKeyShare(bytes32 indexed sarcoId, bytes rawKeyShare);
```
katien, line 14: Renaming all instances of `unwrap` to `publishKeyShare` to make it more clear that archaeologists aren't decrypting the sarcophagus, they're just supplying the share they're responsible for

contracts/facets/ArchaeologistFacet.sol
```diff
@@ -163,58 +155,78 @@ contract ArchaeologistFacet {
...
-
-        // Confirm that the sender is an archaeologist on this sarcophagus
-        if (!LibUtils.archaeologistExistsOnSarc(sarcoId, msg.sender)) {
-            revert LibErrors.ArchaeologistNotOnSarcophagus(msg.sender);
+    /// @param rawKeyShare The keyshare the archaeologist is publishing
+    function publishKeyShare(bytes32 sarcoId, bytes calldata rawKeyShare) external {
+        LibTypes.Sarcophagus storage sarcophagus = s.sarcophagi[sarcoId];
+
+        // Confirm sarcophagus exists
+        if (sarcophagus.resurrectionTime == 0) {
```
katien, line 168: I'll create utilities for existence checks in a future PR

contracts/facets/ArchaeologistFacet.sol
```diff
@@ -163,58 +155,78 @@ contract ArchaeologistFacet {
...
+                s.gracePeriod,
+                block.timestamp
+            );
+        }
+
+        // Confirm tx sender is an archaeologist on the sarcophagus
+        LibTypes.CursedArchaeologist storage cursedArchaeologist = s
+            .sarcophagi[sarcoId]
+            .cursedArchaeologists[msg.sender];
+        if (cursedArchaeologist.doubleHashedKeyShare == 0) {
```
katien, line 200: Not sure if this is worth having a utility for or just something we need to remember, but checking for the default empty state on `doubleHashedKeyShare` (a `bytes32`) and the default empty state on `rawKeyShare` (a `bytes`) is different in a subtle way:
bytes32 we treat like a number and check for a value of zero
`cursedArchaeologist.doubleHashedKeyShare == 0`
bytes we treat like a string and check for empty length
`cursedArchaeologist.rawKeyShare.length != 0`
_checking for length > 0 on a bytes32 will always return true because the bytes32 is initialized with 32 zeroes ☠️_

contracts/facets/EmbalmerFacet.sol
```diff
@@ -22,7 +22,7 @@ contract EmbalmerFacet {
         address[] cursedArchaeologists,
         uint256 totalDiggingFees,
         uint256 createSarcophagusProtocolFees,
-        string[] arweaveTxIds
+        string[2] arweaveTxIds
```
katien, line 25: Typing this as a tuple enforces the validation we were doing to be sure that 2 ids are passed in

contracts/facets/EmbalmerFacet.sol
```diff
@@ -34,267 +34,237 @@ contract EmbalmerFacet {
...
+    /// @param arweaveTxIds ordered pair of arweave tx ids: [sarcophagus payload tx, encrypted key share tx]
+    /// @return The index of the new sarcophagus in sarcophagusIdentifiers
     function createSarcophagus(
         bytes32 sarcoId,
-        LibTypes.SarcophagusMemory memory sarcophagus,
-        LibTypes.SelectedArchaeologistData[] memory selectedArchaeologists,
-        string[] memory arweaveTxIds
+        SarcophagusParams calldata sarcophagusParams,
+        SelectedArchaeologistData[] calldata selectedArchaeologists,
+        string[2] memory arweaveTxIds
```
katien, line 81: Typing this as a tuple enforces the validation we were doing to be sure that 2 ids are passed in

contracts/facets/EmbalmerFacet.sol
```diff
@@ -34,267 +34,237 @@ contract EmbalmerFacet {
...
-            maximumRewrapInterval: sarcophagus.maximumRewrapInterval,
-            arweaveTxIds: arweaveTxIds,
-            embalmer: msg.sender,
-            recipientAddress: sarcophagus.recipient,
-            archaeologists: cursedArchaeologists
-        });
-
-        // Add the identifier to the necessary data structures
+        // update sarcophagus-specific convenience lookup structures
         s.sarcophagusIdentifiers.push(sarcoId);
```
katien, line 188: Are we using `sarcophagusIdentifiers`?

contracts/facets/EmbalmerFacet.sol
```diff
@@ -34,267 +34,237 @@ contract EmbalmerFacet {
...
-            cursedArchaeologists,
+            sarcophagusParams.recipientAddress,
+            sarcophagus.cursedArchaeologistAddresses,
             totalDiggingFees,
             protocolFees,
             arweaveTxIds
         );
 
-        // Return the index of the sarcophagus
+        // return index of sarcophagus in sarcophagusIdentifiers
```
katien, line 209: are we using `sarcophagusIdentifiers`? I didn't see it exposed in viewstatefacet

contracts/facets/EmbalmerFacet.sol
```diff
@@ -303,69 +273,60 @@ contract EmbalmerFacet {
...
-            LibTypes.ArchaeologistStorage memory archaeologistData = LibUtils
-                .getArchaeologist(sarcoId, bondedArchaeologists[i]);
+        // Set resurrection time to infinity
+        sarcophagus.resurrectionTime = 2**256 - 1;
 
+        // for each archaeologist on the sarcophagus, unlock bond and pay digging fees
+        address[] storage archaeologistAddresses = sarcophagus.cursedArchaeologistAddresses;
+        for (uint256 i = 0; i < archaeologistAddresses.length; i++) {
+            // return locked bond to archaeologist
+            LibBonds.freeArchaeologist(sarcoId, archaeologistAddresses[i]);
```
katien, line 322: needs to check for accusals

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -28,185 +28,174 @@ contract ThirdPartyFacet {
...
-        sarco.state = LibTypes.SarcophagusState.Cleaned;
-
-        emit CleanUpSarcophagus(
-            sarcoId,
-            msg.sender,
-            cleanerBondReward,
-            embalmerBondReward
-        );
-    }
+//    function clean(bytes32 sarcoId, address paymentAddress) external {
```
katien, line 31: Commented out the entire function instead of updating it to work with the new data structures because it's getting rethought

contracts/facets/ViewStateFacet.sol
```diff
@@ -242,54 +212,61 @@ contract ViewStateFacet {
...
+        uint256 resurrectionTime;
+        bool isCompromised;
+        string name;
+        uint8 threshold;
+        uint256 maximumRewrapInterval;
+        string[2] arweaveTxIds;
+        address embalmerAddress;
+        address recipientAddress;
+        address[] archaeologistAddresses;
+        uint8 publishedKeyShareCount;
```
katien, line 225: We're not able to return the entire sarcophagus object (with its mapping of cursed archaeologists) because we can't return a dynamic type from a view function. Instead, I added two aggregate properties `publishedKeyShareCount` and `hasLockedBond` on the return type so we'll have the data we need for the dashboard with only one request.

contracts/libraries/LibTypes.sol
```diff
@@ -2,119 +2,38 @@
...
-        uint256 diggingFeesPaid;
-        bytes32 unencryptedShardDoubleHash;
-        bytes unencryptedShard;
+        bytes rawKeyShare;
     }
 
-    // ArchaeologistProfile is used to store archaeologist profile data
     struct ArchaeologistProfile {
-        bool exists;
+        bool exists; // todo: use peerid.length instead of exists
```
katien, line 32: We may be able to get rid of this field and check for an empty peerId instead

package.json
```diff
@@ -33,10 +33,10 @@
   },
   "homepage": "https://github.com/sarcophagus-org/sarcophagus-v2-contracts#readme",
   "devDependencies": {
+    "@nomicfoundation/hardhat-chai-matchers": "^1.0.4",
```
katien, line 36: Migrated us to hardhat chai matchers from waffle https://hardhat.org/hardhat-chai-matchers/docs/migrate-from-waffle

contracts/facets/EmbalmerFacet.sol
```diff
@@ -34,267 +34,237 @@ contract EmbalmerFacet {
...
-            // previous iterations in this loop.
-            if (LibUtils.archaeologistExistsOnSarc(sarcoId, arch.archAddress)) {
-                revert LibErrors.ArchaeologistListNotUnique(
-                    cursedArchaeologists
-                );
+            // confirm archaeologist is registered
+            LibUtils.revertIfArchProfileDoesNotExist(selectedArchaeologists[i].archAddress);
+
+            // Confirm archaeologist isn't already cursed on sarcophagus
+            // todo: may be unnecessary, is cursing an archaeologist twice harming anybody but the caller?
```
DarksightKellar, line 143: Would it not hurt the archaeologist?

contracts/facets/EmbalmerFacet.sol
```diff
@@ -34,267 +34,237 @@ contract EmbalmerFacet {
...
-                sarcophagus.maximumRewrapInterval,
-                sarcophagus.timestamp,
-                arch.diggingFee,
-                arch.v,
-                arch.r,
-                arch.s,
-                arch.archAddress
+                sarcophagusParams.maximumRewrapInterval,
+                sarcophagusParams.creationTime,
+                selectedArchaeologists[i].diggingFee,
```
DarksightKellar, line 156: I'm interested in the difference in cost between having to do this "i-th" lookup every time vs simply accessing a known object in memory. Looks like this change was made because it's actually more efficient?

contracts/facets/EmbalmerFacet.sol
DarksightKellar: We might have considered this at some point before, iirc, and I think the argument was that the archaeologist will want to be able keep track of their earnings on the system. This way they won't have to remember how much of their free bond pool came from their pockets in order to know how much they have earned.

contracts/libraries/LibTypes.sol
DarksightKellar: Agreed

contracts/facets/EmbalmerFacet.sol
katien: It would just result in them getting their bond locked twice and their digging fees paid twice - so equivalent to the archaeologist getting cursed on two sarcophagi with the same digging fees

contracts/facets/ArchaeologistFacet.sol
katien: I will make a utility 😅

contracts/facets/EmbalmerFacet.sol
DarksightKellar: Ah, true.

Although... it wouldn't be what the archaeologist explicitly signed up for. If we give them an interface for checking out their current responsibilities, they'll see 1 sarco, but total locked bond for 2. Probably won't have any way of knowing they'll be paid twice during rewrap/unwrap in the future.

Even if after the initial confusion it becomes a pleasant surprise, it _is_ a "bug" and might be abused in ways we've not even thought of

contracts/facets/ArchaeologistFacet.sol
knoll3: I agree with using the term "unencrypted" since that's how we have it throughout the entire app.

knoll3: Are we going with using the word "share" globally instead of "shard". Technically is should probably be "share" but we have "shard" all over the place.

contracts/facets/EmbalmerFacet.sol
sethhrbek:
```
sarcophagus.cursedArchaeologists[selectedArchaeologists[i].archAddress] = LibTypes
                .CursedArchaeologist({
                    isAccused: false,
                    diggingFee: selectedArchaeologists[i].diggingFee,
                    doubleHashedKeyShare: selectedArchaeologists[i].doubleHashedKeyShare,
                    rawKeyShare: ""
                });
```

It seems like if we didn't check this, then the `cursedArchaeologists` could get out of sync with `cursedArchaeologistAddresses`. Bc you could be cursing with 2 different keyshares on the same arch, but `cursedArchaeologistAddresses` would only save one of them (since its indexed by archAddress)

contracts/facets/EmbalmerFacet.sol
```diff
@@ -34,267 +34,237 @@ contract EmbalmerFacet {
...
+            block.timestamp + sarcophagusParams.maximumRewrapInterval <
+            sarcophagusParams.resurrectionTime
+        ) {
+            revert LibErrors.ResurrectionTimeTooFarInFuture(
+                sarcophagusParams.resurrectionTime,
+                sarcophagusParams.maximumRewrapInterval
+            );
         }
 
+        // todo: we may not need to validate k and n, is the embalmer hurting anybody by themselves by setting incorrect k/n values?
```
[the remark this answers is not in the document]
knoll3: What about the case where the caller calls the contracts directly, bypassing the front end? Do you think we should still validate for that?

knoll3 (approved):

contracts/facets/EmbalmerFacet.sol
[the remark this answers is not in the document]
katien: discussed during standup - we're keeping all validation for now but in a future task around gas optimization we'll drop some on a case by case basis

contracts/facets/EmbalmerFacet.sol
katien: Oh you're right, this would cause problems with the archaeologist not being able to unlock their bonds twice - will remove this todo. So far I've kept all validation in the contracts except the check for arweave tx ids which is getting enforced through typing now. I'm actually going to remove some of these validation todos because I think the possibility of some nonstandard state causing unexpected behavior isn't worth the gas savings with most of these. When we get to the gas cost evaluation task we may revisit this idea.

contracts/facets/EmbalmerFacet.sol
katien: Copying each of those structs from calldata into memory has some cost, so I'd rather just access it where it currently lives in calldata. I couldn't find any clear info on whether or not array indexing has extra cost though, I even asked on a forum after you posted this question because you made me curious. My feeling is that this is the more efficient way to do things, but I'd like to actually do a gas cost evaluation or see if we can get the yul code for this and figure out what's going on under the hood.

contracts/facets/ArchaeologistFacet.sol
katien: In the contracts what we're really trying to differentiate between is the doubleHash of the keyShare and the unprocessed (raw) keyShare, since there's no encrypted key share anywhere. 

I can understand the distinction between encrypted and unencrypted being a lot more important client side though and I think this is a scenario where it's appropriate for the webapp to use `unencryptedKeyShare` because the webapp will have `encryptedKeyShare`s as well. 

As for shares vs shards I think we ought to standardize on shares across all codebases and documentation (and definitely in the UI) to avoid any confusion.

contracts/facets/ArchaeologistFacet.sol
katien: yep! Something like that

PR #103, opened 2022-11-30
Add failures to arch stats

knoll3:
Adds failures to the arch stats in the view state function `getArchaeologistsStatistics`.

Looking at this further, the only way to avoid additional logic in the view state function is to make hundreds of RPC calls from the web app. 
1. First we would need to look up each arch's sarcophagi, an RPC call per arch
2. Second, we would need to make an RPC call per sarcophagus to check it's resurrection time to see if it has failed.

This would be an RPC call for each sarco on each arch, which for my local tests was already reaching the hundreds.

Having this logic in `getArchaeologistsStatistics` requires only 1 RPC call at the cost of adding a nested loop to the function.

### DIFF
--- a/contracts/facets/ViewStateFacet.sol
+++ b/contracts/facets/ViewStateFacet.sol
@@ -207,7 +207,7 @@ contract ViewStateFacet {
             uint256 failures = 0;
 
             // Get arch sarco ids
-            bytes32[] memory sarcoIds = this.getArchaeologistSarcophagi(addresses[i]);
+            bytes32[] storage sarcoIds = s.archaeologistSarcophagi[addresses[i]];
 
             // For each sarco id, if the sarco id is not included in successes and resurrection time
             // has passed, it's a failure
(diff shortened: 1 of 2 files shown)


contracts/facets/ViewStateFacet.sol
```diff
@@ -202,10 +203,29 @@ contract ViewStateFacet {
             memory statsList = new ArchaeologistStatistics[](addresses.length);
 
         for (uint256 i = 0; i < addresses.length; i++) {
+            // Count the failures for the arch
+            uint256 failures = 0;
+
+            // Get arch sarco ids
+            bytes32[] memory sarcoIds = this.getArchaeologistSarcophagi(addresses[i]);
```
katien, line 210: Could we retrieve `s.archaeologistSarcophagi[archaeologist]` directly from appstorage and use a reference type instead of allocating a new `bytes32[]` in memory?

`bytes32[] storage sarcoIds = s.archaeologistSarcophagi[archaeologist]`

Then we might not even need that utility `getArchaeologistSarcophagi` unless it's being used client side

knoll3: Good idea

katien (approved):

PR #109, opened 2022-12-09
DRAFT: Update publishkeyshares to publishprivatekey

knoll3:
Do not merge this. I would just like to get some eyes on it.

- Updated the `publishKeyShare` function to accept a private key and renamed it to `publishPrivateKey`. 
- Added some tests for `publishPrivateKey`.
- Updated the accuse function to receive signatures.


contracts/libraries/LibTypes.sol
```diff
@@ -36,4 +36,10 @@ library LibTypes {
         uint256 freeBond;
         uint256 cursedBond;
     }
+
+    struct Signature {
```
katien, line 40: Think we're going to use this struct outside of thirdpartyfacet? If no, let's move it in there

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -109,28 +112,35 @@ contract ThirdPartyFacet {
...
-            LibTypes.CursedArchaeologist storage accusedArchaeologist = sarcophagus.cursedArchaeologists[accusedArchaeologistAddress];
-
-            // verify the accused archaeologist is cursed on the sarcophagus
-            if (accusedArchaeologist.doubleHashedKeyShare == 0) {
-                revert LibErrors.ArchaeologistNotOnSarcophagus(msg.sender);
+        for (uint256 i = 0; i < signatures.length; i++) {
+            // Recover the archaeologist's address from the signature. Note that the recoverAddress
+            // function will always return a valid address, even if the message does not match the
+            // signature.
+            address accusedArchaeologistAddress = LibUtils.recoverAddress(
```
katien, line 130: This is going to recover the address corresponding to the public key the archaeologist has generated for the sarcophagus, but it's not going to be the same as the archaeologist's address.

We'll need to recover the full public key from the signature, then look up the archaeologist address given that public key in the new convenience mapping we're creating (publicKey=>archaeologistAddress)

PR #115, opened 2022-12-18
110 update cursedarchaeologist to store private and public key

katien:
This PR contains all changes required to support non-layered sarcophagus encryption. The motivation for this refactor and changes are summarized below.

# Original Design
- Sarcophagus data payloads are encrypted once with the recipient key and a second time with a key generated by the embalmer, split, and distributed across several archaeologist nodes
- Nodes publish their key shares at resurrection time allowing anybody to decrypt the outer layer of the sarcophagus payload
- If a node leaks their key share prior to resurrection time, any user with access to the leaked key share can accuse the node of misbehavior and claim a portion of the node's locked bond

The problem is that with this design the embalmer has enough information to immediately issue a valid accusation on all archaeologist nodes. The embalmer has access to all key shares because the embalmer has generated them. There is a financial incentive to become an embalmer specifically for the purpose of accusing archaeologists and stealing their bonds.

# New Design
- Sarcophagus payload is encrypted with a single sarcophagus encryption key instead of a recipient key and an embalmer key
- Sarcophagus encryption key is split via Shamir's secret sharing with each share assigned to an archaeologist node
- Each archaeologist generates a new key for the sarcophagus and supplies the public portion to the embalmer
- Each key share is encrypted once with the recipient key and a second time with the corresponding archaeologist's supplied public key
- Double encrypted key shares are stored on Arweave along with the encrypted sarcophagus payload
- At resurrection time, the archaeologists publish the private keys which can be used by anybody to decrypt the outer layer of encryption on each share of the sarcophagus encryption key
- Only the recipient can finish decrypting the inner layer of encryption on the key shares
- Once the shares are decrypted, the sarcophagus encryption key can be reconstructed and the recipient can access the sarcophagus payload
- The archaeologist will be the only party that ever has access to their private keys, so the malicious embalmer accusal issue will be solved

# New Sarcophagus Creation Flow
1. Embalmer generates a sarcophagus encryption key, this is the only key that is used to encrypt the sarcophagus payload itself
2. Embalmer uses Shamir's secret sharing to split the sarcophagus encryption key into K shares to be assigned to K archaeologists where N shares are required to reconstruct the original key
3. Embalmer encrypts each key share once with the sarcophagus recipient's key
4. Each archaeologist that the embalmer has selected generates a new key for the sarcophagus and sends it to the embalmer
5. Embalmer encrypts each key share a second time with the key from its assigned archaeologist
6. Embalmer uploads the set of double encrypted key shares `[archaeologistKeyEncryption(recipientKeyEncryption(rawKeyShare1))...archaeologistKeyEncryption(recipientKeyEncryption(rawKeyShareK))]` to Arweave along with the encrypted sarcophagus payload
7. Each archaeologist digitally signs the parameters of the sarcophagus (maximum rewrap interval, creation time, digging fees, and the public key that the archaeologist generated for the sarcophagus). This signature allows the contracts to verify that the archaeologist is in possession of the private key required to resurrect the sarcophagus.
8. Embalmer calls `createSarcophagus` on the smart contracts with the archaeologist public keys which are stored on-chain to validate accusals and resurrections

# New Resurrection Flow
1. When an archaeologist detects that a sarcophagus they are tasked with protecting has reached its resurrection time, they publish the corresponding private key
2. Recipient retrieves the encrypted sarcophagus and the double encrypted key shares off of Arweave
3. Recipient decrypts the outer layer of each key share using the private keys that the archaeologists have published
4. Recipient decrypts the inner layer of each key share with their own key and reconstructs the original sarcophagus encryption key
5. Recipient uses the reconstructed sarcophagus encryption key to decrypt the sarcophagus payload

# Accuse
- Accusals occur if an archaeologist has leaked a private key for a sarcophagus they're protecting
- Accuser must be in possession of the leaked private key
- Accuser uses the leaked key to digitally sign the sarcophagus id and a payment address
- Contracts will pay out a portion of the archaeologist's bond to the payment address if the signature can be verified as coming from the private key corresponding to the archaeologist's public key on the sarcophagus with the supplied id
- This pattern prevents the original frontrunning issue because the payment address is baked into the accusal transaction in a way that cannot be altered (if a malicious user attempts to copy the tx and modify the payment address the signature will be invalid)


contracts/facets/EmbalmerFacet.sol
```diff
@@ -73,51 +134,58 @@ contract EmbalmerFacet {
...
     /// @param selectedArchaeologists the archaeologists the embalmer has selected to curse
     /// @param arweaveTxIds ordered pair of arweave tx ids: [sarcophagus payload tx, encrypted key share tx]
-    /// @return The index of the new sarcophagus in sarcophagusIdentifiers
     function createSarcophagus(
         bytes32 sarcoId,
         SarcophagusParams calldata sarcophagusParams,
         SelectedArchaeologistData[] calldata selectedArchaeologists,
         string[2] memory arweaveTxIds
-    ) external returns (uint256) {
+    ) external {
```
sethhrbek, line 142: should add a TODO here to only pass in a single arweave TX ID

contracts/facets/EmbalmerFacet.sol
```diff
@@ -142,23 +210,15 @@ contract EmbalmerFacet {
...
-                sarcophagusParams.creationTime,
-                selectedArchaeologists[i].diggingFee,
-                selectedArchaeologists[i].v,
-                selectedArchaeologists[i].r,
-                selectedArchaeologists[i].s,
-                selectedArchaeologists[i].archAddress
-            );
+            // todo: check convenience structure for public keys that have already been used
+
+            // todo: verify that the sarcophagus parameters have been signed with the private key corresponding to the supplied public key
```
sethhrbek, line 221: Is this going to be extra gas costs to check the specific private key, rather than the arch's registering priv key (b/c we already have address for that key)? If it's not any more expensive, then can go that route, but if it is, then should discuss the tradeoff.

Checking the specific private key => pub key relationship ensures the archaeologist has access to that private key, but the archaeologist has no incentive *not* to provide a public key they don't have the private key for (if they did this, they wouldn't be able to unwrap, losing them $$$ and hurting their reputation).

sethhrbek (approved):

test/facets/embalmer/createSarcophagus.spec.ts
```diff
@@ -150,7 +150,7 @@ describe("EmbalmerFacet.createSarcophagus", () => {
         );
       await expect(tx).to.be.revertedWithCustomError(
         embalmerFacet,
-        `MinShardsGreaterThanArchaeologists`
+        `ThresholdGreaterThanTotalNumberOfArchaeologists`
```
DarksightKellar, line 153: Suggest `ThresholdGreaterThanTotalNumberOfArchaeologists` -> `ThresholdGreaterThanTotalArchaeologists`?

DarksightKellar (approved):

contracts/facets/EmbalmerFacet.sol
```diff
@@ -47,24 +62,70 @@ contract EmbalmerFacet {
...
+    /// @param maximumPermissibleResurrectionTime Resurrection timestamp which is too far in the future
+    error ResurrectionTimeTooFarInFuture(
+        uint256 resurrectionTime,
+        uint256 sarcophagusMaximumRewrapInterval,
+        uint256 maximumPermissibleResurrectionTime
+    );
+
+    error NewResurrectionTimeInPast(uint256 currentTime, uint256 newResurrectionTime);
+
+    error NewResurrectionTimeTooFarInFuture(
```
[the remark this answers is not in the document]
katien: We have another error in the same function that is emitted when a user is trying to rewrap a sarcophagus where the existing resurrection time is in the past so I wanted to make sure we were differentiating in our errors!

PR #116, opened 2022-12-19
DRAFT: Feature/multiple keys update

sethhrbek:
Merges in private key verification
Adds single arweave TX ID (rather than 2)
Adds pub key verification


contracts/facets/EmbalmerFacet.sol
```diff
@@ -104,6 +104,8 @@ contract EmbalmerFacet {
     /// @param resurrectionTime Resurrection timestamp which has already passed
     error ResurrectionTimeInPast(uint256 currentTime, uint256 resurrectionTime);
 
+    error PublicKeyAlreadyUsed();
```
sethhrbek, line 107: this should take a param

contracts/libraries/LibTypes.sol
```diff
@@ -37,4 +37,10 @@ library LibTypes {
         uint256 freeBond;
         uint256 cursedBond;
     }
+
+    struct Signature {
+        uint8 v;
+        bytes32 r;
+        bytes32 s;
+    }
```
sethhrbek, line 45: not sure if this should be here (may exist elsewhere?)

test/facets/archaeologist/publishPrivateKey.spec.ts
```diff
@@ -14,7 +14,7 @@ import {
 
 const { deployments, ethers } = require("hardhat");
 
-describe("bread ArchaeologistFacet.publishPrivateKey", () => {
```
katien, line 17: hah I use this because I hardcode `--grep bread` into the test script in webstorm for easy debugging on specific tests. Sometimes it sneaks into draft prs.

test/libraries/mnemonics.json
```diff
@@ -0,0 +1,102 @@
+[
```
katien, line 1: An easier way to deal with test key generation is
```
 const privateKeys = Array.from(
    { length: totalShares },
    () => "0x" + crypto.randomBytes(32).toString("hex")
  );
  const publicKeys = privateKeys.map(
    (privateKey: string) => new ethers.utils.SigningKey(privateKey).publicKey
  );
  ```

File: contracts/facets/EmbalmerFacet.sol [hunk 5]
```diff
@@ -216,9 +220,20 @@ contract EmbalmerFacet {
                 revert ArchaeologistListContainsDuplicate(selectedArchaeologists[i].archAddress);
             }
 
-            // todo: check convenience structure for public keys that have already been used
+            if(s.publicKeyToArchaeologistAddress[selectedArchaeologists[i].publicKey] != address(0)) {
```
katien, line 223: Need to make sure we have integration tests for new logic like this but I can add that on the other branch

PR #117, opened 2022-12-21
114 write signature verification utility

knoll3:
Adds a utility that verifies a signature using a public key. Rebased on main.

### DIFF
--- a/contracts/libraries/LibUtils.sol
+++ b/contracts/libraries/LibUtils.sol
@@ -59,6 +59,33 @@ library LibUtils {
         }
     }
 
+    /// @notice Verifies that a signature belongs to a given public key
+    /// @param message the message that was signed
+    /// @param publicKey the public key that was used to sign the message
+    /// @param v signature element
+    /// @param r signature element
+    /// @param s signature element
+    /// @return true if the signature is valid, false otherwise
+    function verifySignature(
+        bytes calldata message,
+        bytes calldata publicKey,
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) internal pure returns (bool) {
+        bytes32 messageHash = keccak256(
+            abi.encodePacked("\x19Ethereum Signed Message:\n32", keccak256(message))
+        );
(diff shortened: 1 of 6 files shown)


contracts/libraries/LibUtils.sol
```diff
@@ -59,6 +59,33 @@ library LibUtils {
...
     }
 
+    /// @notice Verifies that a signature belongs to a given public key
+    /// @param message the message that was signed
+    /// @param publicKey the public key that was used to sign the message
+    /// @param v signature element
+    /// @param r signature element
+    /// @param s signature element
+    /// @return true if the signature is valid, false otherwise
+    function verifySignature(
```
katien, line 69: Let's give this a more specific name like `verifyAccusalSignature` or `verifySignatureIsFromPublicKey` if there's a possibility we'll be using this utility elsewhere. We've already got another signature verification utility in this file so things could get confusing

contracts/libraries/LibUtils.sol
```diff
@@ -59,6 +59,33 @@ library LibUtils {
...
+        bytes calldata message,
+        bytes calldata publicKey,
```
katien, line 71: Let's change publicKey to `storage` because we'll always be reading the value from storage on the cursed archaeologist

contracts/libraries/LibUtils.sol
```diff
@@ -59,6 +59,33 @@ library LibUtils {
...
+        uint8 v,
+        bytes32 r,
+        bytes32 s
+    ) internal pure returns (bool) {
+        bytes32 messageHash = keccak256(
+            abi.encodePacked("\x19Ethereum Signed Message:\n32", keccak256(message))
+        );
+        // Use ecrecover to get the address that signed the message (the signer is not the archaeologist)
+        address recoveredAddress = ecrecover(messageHash, v, r, s);
```
katien, line 80: how about we call these `signingAddress` and `publicKeyAddress` for extra clarity

katien (approved):

PR #121, opened 2023-02-09
Digging fees per second update

DarksightKellar:
## Overview

This PR updates the contracts to work with digging fees per second instead a constant digging fee value.

Archaeologists will set a minimum digging fee per second that they would like to earn for being cursed on a sarcophagus. 

Once an archaeologist is bound to a sarcophagus, they are contractually bound to earn the agreed upon digging fees per second for the the lifetime of the sarcophagus, even if they change their minimum on their profile.

Payouts are calculated based on how much time interval exists between the agreed upon resurrection time and either: 
 - the creation time for new sarcophagi
 - the previous rewrap time for a sarcophagus being rewrapped (in which case the *new* resurrection time is the upper bound of the interval).

Archaeologist bonds are similarly locked when being cursed, in order to keep a 1-to-1 ratio of expected payout to locked bond.

## Refactoring done in this PR:
 - `LibBonds.sol` refactored, mostly to remove redundancy and rename functions more descriptively.
 - A new `LibBonds.curseArchaeologist` now handles all operations toward the action of "cursing" an archaeologist. Code that was doing this in `EmbalmerFacet.createSarcophagus` has been accordingly moved.

 - `LibBonds.freeArchaeologist` has been updated to internally calculate digging fees due and add this to the arch rewards.
- A side effect of the above update is that for `bury`, the resurrection time cannot be updated to `MaxUInt256` before the call to `freeArchaeologist`, as reward payout depends on the sarcophagus `resurrectionTime`. I considered having `freeArchaologist` accept this interval instead, but decided against it because it would mean deferring this calculation to its other callers. **(Need to consider potential security implications of this order of execution)**

- Moved `revert SenderNotEmbalmerOrAdmin(msg.sender);` higher up the logic -- no point checking anything else if the sender is unauthorised anyways.

---

Tests have also been updated to account for these updates -- all passing. I have not however considered if more tests need to added or not, to specifically test the new time-based rewarding+bonding mechanism.

### DIFF
--- a/contracts/facets/EmbalmerFacet.sol
+++ b/contracts/facets/EmbalmerFacet.sol
@@ -11,7 +11,6 @@ import {LibUtils} from "../libraries/LibUtils.sol";
 import "hardhat/console.sol";
 
 contract EmbalmerFacet {
-
     /// @notice Emitted when a sarcophagus is created
     /// @param sarcoId Id of the new sarcophagus
     /// @param name Name of the new sarcophagus
@@ -53,7 +52,7 @@ contract EmbalmerFacet {
     /// @notice Parameters of a sarcophagus, supplied during sarcophagus creation
     struct SarcophagusParams {
         string name;
-        // highest rewrap interval bonded archaeologists have agreed to accept for lifetime of sarcophagus
+        // highest rewrap interval cursed archaeologists have agreed to accept for lifetime of sarcophagus
         uint256 maximumRewrapInterval;
         address recipientAddress;
         uint256 resurrectionTime;
@@ -65,8 +64,7 @@ contract EmbalmerFacet {
     struct CurseParams {
         bytes publicKey;
         address archAddress;
-        // diggingFee archaeologist has agreed to receive on sarcophagus for its entire lifetime
-        uint256 diggingFee;
+        uint256 diggingFeePerSecond;
         uint8 v;
         bytes32 r;
         bytes32 s;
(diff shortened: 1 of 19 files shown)


contracts/facets/EmbalmerFacet.sol
```diff
@@ -177,16 +178,16 @@ contract EmbalmerFacet {
...
             revert NoArchaeologistsProvided();
         }
-        // Confirm that threshold is greater than 0
+
         if (sarcophagusParams.threshold == 0) {
             revert ThresholdCannotBeZero();
         }
-        // Confirm that threshold is less than or equal to the number of archaeologists
-        // (k <= n in a shamir secret sharing scheme)
+
```
knoll3, line 189: Is there a missing comment here?

contracts/facets/EmbalmerFacet.sol
```diff
@@ -336,7 +322,7 @@ contract EmbalmerFacet {
             );
         }
 
-        // track total digging fees across all archaeologists on the sarcophagus
+        // track total digging fees to be paid by embalmber across all archaeologists on the sarcophagus
```
knoll3, line 325: Change "embalmber" to "embalmer"

contracts/libraries/LibBonds.sol
```diff
@@ -40,77 +42,77 @@ library LibBonds {
...
-    /// @param archaeologist The address of the archaeologist
-    /// @param amount The amount to unlock
-    function unlockBond(address archaeologist, uint256 amount) internal {
-        // Decrease the cursed bond amount
-        decreaseCursedBond(archaeologist, amount);
+        s.archaeologistSarcophagi[archaeologist.archAddress].push(sarcoId);
 
-        // Increase the free bond amount
-        increaseFreeBond(archaeologist, amount);
+        return diggingFeesDue;
```
knoll3, line 97: Is there a reason to return `diggingFeesDue`?

knoll3 (approved):
Good work. I don't see anything that stands out.

contracts/libraries/LibBonds.sol
DarksightKellar: Yep -- digging fees calculation has been moved to this curse function, so the result is returned so that `createSarcophagus` can track the *total* digging fees due all archaeologists

contracts/facets/EmbalmerFacet.sol
DarksightKellar: 👍🏾

contracts/facets/EmbalmerFacet.sol
DarksightKellar: Nope. This comment is actually probably unnecessary altogether. I'll update it though to read a little less abruptly

PR #125, opened 2023-02-15
DRAFT: Locked bond ratio

sethhrbek:
Adds cursed bond percentage, which determines the ratio that archaeologist must lock up when accepting curses.

It defaults to a 1:1 ratio (`cursedBondRatio = 100`).

On rewraps, if an archaeologist must lock up further bond, bond will be taken from rewards from the previous cycle of the sarcophagus.


contracts/libraries/LibBonds.sol
```diff
@@ -96,11 +99,22 @@ library LibBonds {
...
+
+        decreaseCursedBond(archaeologistAddress, cursedBondAmount);
+        s.archaeologistProfiles[archaeologistAddress].freeBond += cursedBondAmount;
+        s.archaeologistRewards[archaeologistAddress] += diggingFeeAmount;
+    }
+
+    /// @notice Calculates the protocol fees to be taken from the embalmer.
+    /// @param totalDiggingFees to be paid. Protocol fee is a percentage of this
+    /// @return The protocol fees amount
+    function calculateCursedBond(uint256 diggingFees) internal view returns (uint256) {
```
DarksightKellar, line 115: Ah, `calculateCursedBond`, we meet again lol

contracts/libraries/LibBonds.sol
```diff
@@ -77,8 +77,11 @@ library LibBonds {
         uint256 diggingFeesDue = archaeologist.diggingFeePerSecond *
             (sarcophagus.resurrectionTime - sarcophagus.previousRewrapTime);
 
-        decreaseFreeBond(archaeologist.archAddress, diggingFeesDue);
-        s.archaeologistProfiles[archaeologist.archAddress].cursedBond += diggingFeesDue;
+        // Use cursed bond percentage to determine how much bond to lock up
+        uint256 bondToCurse = (diggingFeesDue * s.cursedBondPercentage) / 100;
```
DarksightKellar, line 81: I'm sure this was the plan already, but just for reference so it doesn't somehow slip through the cracks, the returning `calculateCursedBond` function can be used for this (and other) cursedBond determining logic.

sethhrbek: @DarksightKellar -- I need to run some gas comparisons. If it's cheaper to run it inline like this, then I might remove the `calculateCursedBond` all together

contracts/facets/ThirdPartyFacet.sol
```diff
@@ -314,49 +317,49 @@ contract ThirdPartyFacet {
...
 
-        // Reward caller, and reimburse embalmer
-        (uint256 accuserBondReward, uint256 embalmerBondReward) = _distributeLoot(
-            paymentAddress,
-            sarcophagus,
-            totalDiggingFees
+        uint256 halfTotalCursedBond = totalCursedBond / 2;
+
+        // transfer the cursed half, plus the current digging fees, to the embalmer
+        s.sarcoToken.transfer(
```
DarksightKellar, line 323: Looks like `_distributeLoot` can safely be removed

contracts/libraries/LibBonds.sol
DarksightKellar: Yea now that you mention it, I have asked this question before (function calls vs inline statements), and the gas comparison I did showed inline was marginally but not-negligibly cheaper. I'll let you run the comparisons on *this* code to confirm, in which case removing `calculateCursedBond` might be the play

PR #126, opened 2023-02-15
Modify locked bond on rewrap

knoll3:
Modifies the rewrap function to change the arch locked bond when the interval changes and adds tests for this functionality. This PR assumes a 1:1 ratio between digging fees and locked bond, although it shouldn't be difficult to make the necessary changes when we change that 1:1 ratio.


DarksightKellar (approved):
Yup. As good as it gets. There's a leftover `.only` in the tests but i'll go ahead and push a commit to remove it

PR #135, opened 2023-03-27
Update events

DarksightKellar:


contracts/facets/ArchaeologistFacet.sol
```diff
@@ -281,6 +285,6 @@ contract ArchaeologistFacet {
         // Save the successful sarcophagus against the archaeologist
         s.archaeologistSuccesses[msg.sender].push(sarcoId);
 
-        emit PublishPrivateKey(sarcoId, privateKey);
+        emit PublishPrivateKey(sarcoId, privateKey, msg.sender);
```
sethhrbek, line 288: @DarksightKellar -- thinking we should remove the `archaeologistSuccesses` array from storage in this PR, since that data can all be retrieved from event data.

DarksightKellar: I think we should remove it _after_ subgraph repo is ready -- so a swap rather than a preemptive removal

PR #145, opened 2023-04-17
added new events

0xbhagi:


0xbhagi:
> @0xbhagi Just following up on comments from @sethhrbek -- Events are not meant to track state, so the single `UpdateArchaeologist` event is all that is needed, with just the line change you pointed out in your issue.
> 
> Also, we're able to index data from events even in the past, so there's no need for dedicated "before" and "after" events. Thanks!

i thought there will be any use case for state track , now as per your reference single 'UpdateArchaeologist' event will be enough.

DarksightKellar (approved):

